PR #12, opened 2021-12-20
Fix version of crate x86_64 to "0.14.6"

gaojiaqi7:
Latest version "0.14.7" of x86_64 fails to build due to error[E0658]:
panicking in constant functions is unstable.


jyao1:
Merge as work-around to pass build.

PR #22, opened 2021-12-27
Lock version of crate bootloader

gaojiaqi7:
Latest version of bootloader depends on the version 0.14.7 of x86_64,
so lock the version for now.


jyao1:
Fix CI issue, merge it.

PR #25, opened 2021-12-29
rust-test-payload: Refine the test-payload

mxu9:
In rust-test-payload there are multi test cases, such as Tdinfo, Acpi,
Event log, Multi-processor, etc. This commit refines the rust-test-payload
so that developer can focuse on the test case itself.

There are below major changes.
1. trait TestCase
  TestCase defines 5 interfaces which should be implemented by the actual
  test case.
  - setup(&mut self)
  - run(&mut self)
  - teardown(&mut self)
  - get_name(&mut self) -> String
  - get_result(&mut self) -> TestResult

2. struct TestSuite
  testsuite is the member of TestSuite which holds the test cases which
  implementes the trait of TestCase.

3. enum TestResult
  Done means the test case is finished. Error means there is something
  wrong when the case run.
  Note: External parse script takes the responsibility to determines if
  the test case pass or fail. Because test case itself doesn't have the
  enough knowlodge. For example, the guest is configured with 8 CPU but
  Tdinfo call returns 4 CPU. payload itself doesn't know it is correct or
  not.

4. struct Tdinfo
  Tdinfo is the actual test case which implements trait of TestCase.

If some more test case is to be added, Tdinfo can be the example. After
a new case is defined, it can be created and added into TestSuite.
    let mut tdinfo = Tdinfo {
        name: String::from("Tdinfo"),
        hob: hob,
        result: TestResult::Error
    };
    ts.testsuite.push(Box::new(tdinfo));

Below is the sample of test-payload log.

INFO - Start to run tests.
INFO - ---------------------------------------------
INFO - [Test: Tdinfo]                    <-- test case name
INFO - td_info data addr: 0x7f7ffe38     <-- below is the log by tdinfo
INFO - gpaw - 52
INFO - attributes - 9223372036854775809
INFO - max_vcpus - 8
INFO - num_vcpus - 8
INFO - rsvd - [0, 0, 0]
INFO - [Test: Tdinfo] - Done             <-- test is done
INFO - ---------------------------------------------
INFO - Total: 1, Done: 1, Failed: 0      <-- summary

Signed-off-by: Min Xu <min.m.xu@intel.com>

### DIFF
--- a/rust-test-payload/src/lib.rs
+++ b/rust-test-payload/src/lib.rs
@@ -1,0 +1,55 @@
+// Copyright (c) 2021 Intel Corporation
+//
+// SPDX-License-Identifier: BSD-2-Clause-Patent
+
+#![no_std]
+extern crate alloc;
+
+use alloc::boxed::Box;
+use alloc::string::String;
+use alloc::vec::Vec;
+
+#[derive(Copy, Clone)]
+pub enum TestResult {
+    Done,
+    Error,
+}
+
+pub trait TestCase {
+    fn setup(&mut self);
+    fn run(&mut self);
+    fn teardown(&mut self);
+    fn get_name(&mut self) -> String;
+    fn get_result(&mut self) -> TestResult;
+}
+
+pub struct TestSuite {
+    pub testsuite: Vec<Box<dyn TestCase>>,
+    pub done_cases: u32,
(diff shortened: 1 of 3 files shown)


jyao1:
Please add license header

mxu9:
License header is added. Please help to review. Thanks

PR #26, opened 2021-12-29
rust-tdshim: fix MP wakeup issue

gaojiaqi7:
This is a regression introduced by MP memory acceptance.

Value of register r9 is used to store the cpu index of CPU, and it could
be changed by function assigned by BSP, so save the value into stack
before calling the function.

Correct the cpu exiting count value checked by BSP, APs can jump to
.check_command directly rather than .do_wait_loop.


jyao1:
Please add more test: 1) increase CPU number, 2) add stress test to wake up AP hundred times.

jyao1:
Merged by mistake.
Please fix the issue on stress test and resubmit PR.

PR #27, opened 2021-12-31
Link FADT and DSDT from VMM

gaojiaqi7:
The Fixed ACPI Description Table (FADT) should be the first table
pointed by Extended System Description Table (XSDT). The Differentiated
System Description Table (DSDT) is refered by the FADT table.

The offset of FADT and DSDT in ACPI memory will be stored and used when
generating XSDT and RSDP.


rust-tdshim/src/acpi.rs
```diff
@@ -135,8 +138,36 @@ impl AcpiTables {
...
+
+        // The Fixed ACPI Description Table (FADT) should
+        // always be the first table that XSDT points to.
+        //
+        if let Some(fadt) = self.fadt {
+            let fadt_header = GenericSdtHeader::read_from(
+                &self.acpi_memory[fadt..fadt + size_of::<GenericSdtHeader>()],
+            )
+            .unwrap();
+            if fadt_header.revision == ACPI_FADT_MAJOR_VERSION {
```
jyao1, line 150: Why the code checks FADT version?
DSDT should be included in FADT in any version.

gaojiaqi7: Agree, it is a mistake. I have updated the patch. DSDT table will be linked to DSDT field(4-bytes) of FADT since ACPI memory of td-layout is below 4G.

PR #29, opened 2022-01-05
rust-tdshim: fix MP wakeup issue

gaojiaqi7:


gaojiaqi7:
This patch has passed the multi CPU and multi cycle stress test

rust-tdshim/src/mp.rs
```diff
@@ -157,55 +159,53 @@ fn parallel_accept_memory(apic_id: u64) {
...
-    let mail_box = MailBox::new(get_mem_slice_mut(SliceType::MailBox));
-    log::info!("Waiting for APs to exit...\n");
-    loop {
-        let cpu_exit = mail_box.cpu_exiting();
-
-        if cpu_exit == 1 {
-            log::info!("All APs has exited\n");
-            break;
+    for i in 1..cpu_num {
+        mail_box.set_apic_id(i);
```
jodh-intel, line 166: What about APIC ID 0? Is that invalid? If so, shouldn't `set_apic_id()` check this and return a `Result<()>`? Either way, a comment explaining why this is 1-based and not zero-based would help imho.

rust-tdshim/src/mp.rs
```diff
@@ -157,55 +159,53 @@ fn parallel_accept_memory(apic_id: u64) {
...
+        mail_box.set_command(MP_WAKEUP_COMMAND_AVAILABLE);
+
+        loop {
+            if mail_box.apic_id() == MAILBOX_APICID_INVALID {
```
jodh-intel, line 170: But if APIC 0 is invalid, should `MAILBOX_APICID_INVALID` be renamed to `MAILBOX_APICID_NOT_SET`?

rust-tdshim/src/mp.rs
```diff
@@ -157,55 +159,53 @@ fn parallel_accept_memory(apic_id: u64) {
...
 
     loop {
-        let wakeup_apic_id = mail_box.apid_id();
-        if wakeup_apic_id == MAILBOX_APICID_INVALID {
+        if mail_box.apic_id() == MAILBOX_APICID_INVALID {
+            mail_box.set_command(MP_WAKEUP_COMMAND_NOOP);
             log::info!("Successfully wakeup AP #{}\n", apic_id);
             break;
+        } else {
+            cpu_pause();
```
jodh-intel, line 195: This code is duplicated in `wait_for_ap_arrive()`. It might be worth creating a new function for this loop that both functions call to remove the risk of future maintenance issues...?

rust-tdshim/src/mp.rs
```diff
@@ -157,55 +159,53 @@ fn parallel_accept_memory(apic_id: u64) {
...
 
-pub fn mp_accept_memory_resource_range(cpu_num: u32, address: u64, size: u64) {
+pub fn mp_accept_memory_resource_range(mut cpu_num: u32, address: u64, size: u64) {
     log::info!(
         "mp_accept_memory_resource_range: 0x{:x} - 0x{:x} ... (wait for seconds)\n",
         address,
         size
     );
 
+    cpu_num = if cpu_num > 16 { 16 } else { cpu_num };
```
jodh-intel, line 207: Why are we limiting the cpu count to 16? What if my system has >16? If this is some sort of internal limit, please create a top-level `const` with details of this restriction (including links to canonical reference material if possible).

jodh-intel (commented):

rust-tdshim/src/mp.rs
gaojiaqi7: @jodh-intel Thanks for your suggestions. The 0 is bootstrap processor (BSP) which is running this code to wakeup other vCPUs. I will add a comment to explain this.

rust-tdshim/src/mp.rs
gaojiaqi7: This variable refers to the cpu count that will be used to do the memory acceptance, the stacks used for application processors (APs) to run rust function are allocated from temp heap of td-shim which is limited. Besides, it is not necessary to use all the APs to do this work if there's too many.
Will create a `const `and explain the reason, thanks.

rust-tdshim/src/mp.rs
gaojiaqi7: Agree, will create a new function to do the common work.

rust-tdshim/src/mp.rs
```diff
@@ -224,23 +231,23 @@ pub fn mp_accept_memory_resource_range(cpu_num: u32, address: u64, size: u64) {
...
-    mail_box.set_fw_arg(1, cpu_num as u64);
+    mail_box.set_fw_arg(1, active_ap_cnt as u64 + 1);
     mail_box.set_fw_arg(2, address + align_low);
     mail_box.set_fw_arg(3, address + size);
 
     if major_part > 0 {
-        for i in 1..cpu_num {
-            let ap_stack = stacks.as_ptr() as u64 + i as u64 * 0x800;
-
+        // 0 is the bootstrap processor running this code
```
jodh-intel, line 248: Thanks. There are a couple of other occurences of `1..` so for maximum safety, you could do something like this to encapsulate the hard-coded offset:

```rust
use std::ops::Range;

fn make_apic_range(end: usize) -> Range<usize> {
    // 0 is the bootstrap processor running this code
    let start = 1;

    Range { start, end }
}

// ...

for i in make_apic_range(cpu_num) {
    // ...
}

for i in make_apic_range(active_ap_cnt) {
    // ...
}

for i in make_apic_range(ap_num) {
    // ...
}
```

jodh-intel (approved):
Thanks @gaojiaqi7.

rust-tdshim/src/mp.rs
gaojiaqi7: It's a good idea, thanks for your advice😊

Issue #9, opened 2021-12-17
Payload type detection


sboeuf:
I have a question about TD Payload related to what the specification states:
```
If the VMM does not provide the payload image or the VMM does not know
the payload image format, the VMM shall not create this payload info HOB.
The TdShim must know the payload format and transfer to the payload in
right way. By default, PayloadImageTypeExecutablePayload is used.
```
Assuming the TD Shim has the ability to identify the payload type, why don't we always rely on the TD Shim to do so? I don't think we should expect the VMM to identify the payload type, especially since it's not doing anything with it other than passing it to the guest.

jyao1:
You are right. If TD shim has ability to identity, then there is no need to let VMM pass the info.

For example, the VMM does not tell TD shim the detail of PayloadImageTypeExecutablePayload (PE/COFF image, or ELF image), the TD shim can detect it is PE/COFF or ELF.

But if TD shim cannot detect, then TD shim relies on VMM to pass the info.

What is your concern? or your proposal?

sboeuf:
Well my understanding from the spec is that TD Shim will __always__ be able to detect the payload type. I'm suggesting that if that's the case, then there's no point in having `HOB_PAYLOAD_INFO_TABLE` at all. The TD Shim should already know where the Payload is located since it provides the information through the TDVF sections already.
Basically I'm proposing some simplification of the spec, by removing the need for the VMM to provide `HOB_PAYLOAD_INFO_TABLE`.

jyao1:
> TD Shim will **always** be able to detect the payload type.

I doubt if this is TRUE.
How TD shim know if a binary blob is PayloadImageTypeBzImage or PayloadImageTypeRawVmLinux ?

sboeuf:
> > TD Shim will **always** be able to detect the payload type.
> 
> I doubt if this is TRUE. 

Oh okay then I guess I read the spec the wrong way. Maybe something like the following would bring some clarity:
```
The VMM must always provide the image format to the TD Shim through
the HOB because the TD Shim might not be able to detect the payload
format.
```

> How does TD shim know if a binary blob is PayloadImageTypeBzImage or PayloadImageTypeRawVmLinux ?

Well the same way you're suggesting the VMM will need to have some detection code, the TD Shim could have the same kind of detection mechanism.

But more than that, since the VMM is not the one preparing the payload and the boot parameters, it could be completely opaque and simply let the TD Shim do the detection.

jyao1:
[...]

PR #32, opened 2022-01-17
[RFC] Detect payload type

sboeuf:
Looking for review and feedback on this PR which adds the support for payload type detection.
It solves issue #9

/cc @jyao1 @rbradford

### DIFF
--- a/rust-tdshim/src/main.rs
+++ b/rust-tdshim/src/main.rs
@@ -145,7 +145,6 @@ fn log_hob_list(hob_list: &[u8], td_event_log: &mut tcg::TdEventLog) {
 
 #[derive(Default, Clone, Copy, Pread, Pwrite)]
 pub struct PayloadInfo {
-    pub image_type: u32,
     pub entry_point: u64,
 }
 
@@ -364,9 +363,7 @@ pub extern "win64" fn _start(
 
     mem.setup_paging();
 
-    if let Some(hob) = hob_lib::get_next_extension_guid_hob(hob_list, &HOB_KERNEL_INFO_GUID) {
-        let kernel_info = hob_lib::get_guid_data(hob).unwrap();
-        let vmm_kernel = kernel_info.pread::<PayloadInfo>(0).unwrap();
+    if hob_lib::get_next_extension_guid_hob(hob_list, &HOB_KERNEL_INFO_GUID).is_some() {
         let payload = memslice::get_mem_slice_mut(memslice::SliceType::Payload);
 
         match detect_payload_type(payload) {
--- a/rust-tdshim/src/main.rs
+++ b/rust-tdshim/src/main.rs
@@ -48,6 +48,7 @@ use core::panic::PanicInfo;
 use core::ffi::c_void;
 
 use crate::e820::create_e820_entries;
+use crate::linux::kernel_param::{Elf64Header, SetupHeader};
 use crate::memory::Memory;
 use crate::memslice::SliceType;
 use crate::verify::PayloadVerifier;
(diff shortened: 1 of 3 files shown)


rbradford:
I think you should also update the spec too.

sboeuf:
>I think you should also update the spec too.

I agree and I will once I'll have received all the feedback.

rust-tdshim/src/main.rs
```diff
@@ -156,6 +156,40 @@ const HOB_KERNEL_INFO_GUID: [u8; 16] = [
...
+
+    if elf64_header.e_ident[0] == 0x7f
+        && elf64_header.e_ident[1] == 0x45
+        && elf64_header.e_ident[2] == 0x4c
+        && elf64_header.e_ident[3] == 0x46
+        && elf64_header.e_ident[5] == 0x1
+        && elf64_header.e_phentsize == 0x38
+        && elf64_header.e_phoff >= 0x40
+    {
+        return PayloadType::RawVmLinux;
```
jyao1, line 178: Why you use this to detect RawVmLinux?
It is incorrect, because a raw image may strip the elf header.

sboeuf: Well the `vmlinux` generated by the kernel build always have the ELF header, that's why I assumed this would be the correct approach. When/How do you get a `vmlinux` without the ELF header?

jyao1: A VMM implementation can do relocation and strip the ELF header.
There is no need to pass the full header, if no load or relocation is required by TdShim.

sboeuf: A few questions regarding TdShim:
1. Why does the spec decided to support both `vmlinux` and `bzImage`? `bzImage` has a well defined boot protocol and it should be enough for what TdShim is designed for.
2. Why do you expect the VMM to do relocation and strip the header in the first place? If the spec wasn't accepting this, we would know for sure what `vmlinux` would look like and where to find it.
3. Since the current TdShim implementation doesn't support the `vmlinux` case, don't you think it'd be easier to update the spec and only expect `bzImage`?
4. Is there a QEMU implementation somewhere? I'm asking because I'm curious what the QEMU community thinks about this part of the TdShim spec.

jyao1:
@jiangliu, would you please comment on `RawVmlinux`?

sboeuf:
ping @jiangliu

c3d (commented):
Since this is an RFC, not going to approve, but I like the general idea.

jyao1:
> I am not personally aware of any relocation in the case of qemu that would strip the header. 

This is not about qemu. This is requested from alibaba's hypervisor. @jiangliu, would you please comment?

jiangliu:
> > I am not personally aware of any relocation in the case of qemu that would strip the header.
> 
> This is not about qemu. This is requested from alibaba's hypervisor. @jiangliu, would you please comment?

Sorry for slow response, I'm go over the td-shim code and is near to the loader now:)
Will give feedback soon:)

rust-tdshim/src/main.rs
jiangliu: Sorry for the slow response and now eventually I reached this part:) @sboeuf @rbradford 
Current there are two ways for the td-shim to load and execute the target linux kernel.
1) The payload area acts as a standard firmware volume and store the target linux kernel as a file in this FV. And [ipl:: find_and_report_entry_point](https://github.com/confidential-containers/td-shim/blob/staging/rust-tdshim/src/ipl.rs#L25) tries to parse the binary. Currently PE32+ and ELF binary are supported.
2) A Kernel Info GUID Hob Entry. When designing the kernel info hob entry, it targets to support raw binary, so it contains an "entrypoint" for td-shim to directly jump to the entry point.

On the other hand, we do have pre-processed binary linux kernel images, so the vmm could directly mmaping the binary file and jump to the entry point. That's why we add the kernel info hob entry to td-shim spec.

slp: > On the other hand, we do have pre-processed binary linux kernel images, so the vmm could directly mmaping the binary file and jump to the entry point. That's why we add the kernel info hob entry to td-shim spec.

libkrun does the exact same trick (I wonder who got it first ;-P). The idea is to prepare the kernel image to set it up in a binary state identical to what the kernel expects to have in memory once it reaches the entry point. That's less work for the guest, and also allows directly sharing the kernel image among multiple instances (by mmap'ing the prepared kernel image, as @jiangliu pointed out).

I'm not sure which name should be given to a kernel image like this. `RawVmlinux` seems a bit confusing to me, but perhaps some docs/comments would do the trick.

jiangliu: So there are two possible ways forward.
1) extend ipl:: find_and_report_entry_point() to support bzImage.
2) use PayloadInfo::image_type to distinguish pre-processed raw binary or normal linux kernel images.

I'm ok with solution 2 above and hope we could keep the image_type field, so we could have image type for:
1) elf kernel
2) bzImage
3) raw binary
4) auto-detect file type

jiangliu:
@sboeuf @c3d  @jyao1 How about this proposal to  enhance the td-shim spec?
https://github.com/confidential-containers/td-shim/pull/79

jyao1:
no discussion in 7 days. Close it now.

If there is concern, please open a new issue or PR.

PR #40, opened 2022-01-26
refactor rust-paging crate

jiangliu:
Refactor the rust-paging crate by:
1) remove unused code from frame.rs
2) get rid of bitmap-allocator
3) refine doc

PR #43, opened 2022-01-28
Split integration test out of rust-paging

jiangliu:
Split integration test out of rust-paging, so we will have a clean dependency for production code. All test related dependences have been moved into tests/rust-test-paging.

This patch is on top of #40 and #41, and the last three patches are new.


jyao1:
good isolation.

jiangliu:
> good isolation.

I'm doing the same isolation for rust-td-payload too:)

PR #45, opened 2022-01-30
Refactor tdx-payload and related unit test cases

jiangliu:
Refactor tdx-payload and related unit test cases.


rust-tdshim/src/mp.rs
```diff
@@ -148,7 +148,8 @@ fn wait_for_ap_responde(mail_box: &mut MailBox) {
 
 // Wait for APs to arrive by checking if they are available
 fn wait_for_ap_arrive(ap_num: u32) {
-    let mut mail_box = MailBox::new(get_mem_slice_mut(SliceType::MailBox));
+    // TODO: Safe because we are the consumer?
```
jyao1, line 151: The mail_box is designed for BSP<->AP communication.
I would say:
1) The consumers for mail_box memory are BSP and APs.
2) The consumers of mail_box data structure are in mp.rs.

Maybe we can clarify that in comments, then we can remove TODO?

rust-tdshim/src/mp.rs
```diff
@@ -189,7 +191,8 @@ fn td_accept_pages(address: u64, pages: u64, page_size: u64) {
 }
 
 fn parallel_accept_memory(apic_id: u64) {
-    let mail_box = MailBox::new(get_mem_slice_mut(SliceType::MailBox));
+    // TODO: safe because we are the only user?
```
jyao1, line 194: What is difference between consumer and user?

I feel a better way is to let parallel_accept_memory() calls MP API.

We may file an issue and do that later.

rust-tdshim/src/mp.rs
jiangliu: Sorry, TODO is marker for me so I could pay attention to these parts when reviewing rust-tdshim:)

rust-tdshim/src/mp.rs
jiangliu: It's actually the same thing, just a marker for myself:)

rust-tdshim/src/mp.rs
jyao1: That is fine. Do you want to keep TODO and check in? Or do you want to post additional patch remove them?

PR #49, opened 2022-02-02
Setup development framework

jiangliu:
Setup development framework with Makefiles and github actions.


.github/workflows/deny.yml
```diff
@@ -0,0 +1,20 @@
+name: cargo-deny
+on: [push, pull_request]
+jobs:
+  cargo-deny:
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        checks:
+          # - advisories
```
c3d, line 9: Just curious why this is disabled.

c3d (approved):
Thanks @jiangliu. Looks good to me. I would add a couple of comments (maybe just reword the commit message) explaining why you disabled some checks and why we need a daily rust toolchain.

sameo (approved):

.github/workflows/deny.yml
jiangliu: This is from the original staging branch.
Actually I hope to enable checking for license:)

PR #50, opened 2022-02-02
dev-container: add Dockerfile to build a dev container

jiangliu:
Add Dockerfile to build a container image for td-shim development
environment.

### DIFF
--- a/devtools/dev_container/Dockerfile
+++ b/devtools/dev_container/Dockerfile
@@ -1,0 +1,24 @@
+FROM ubuntu:20.04
+
+# Adding rust binaries to PATH.
+ENV PATH="$PATH:/root/.cargo/bin"
+ENV CC=clang
+ENV AR=llvm-ar
+
+# Install all required packages in one go to optimize the image
+# https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run
+# DEBIAN_FRONTEND is set for tzdata.
+RUN apt-get update && \
+    DEBIAN_FRONTEND="noninteractive" apt-get install --no-install-recommends -y \
+    build-essential ca-certificates curl gcc git libssl-dev pkg-config ssh \
+    clang llvm nasm \
+    # cleanup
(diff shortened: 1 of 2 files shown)


devtools/dev_container/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+# https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run
+# DEBIAN_FRONTEND is set for tzdata.
+RUN apt-get update && \
+    DEBIAN_FRONTEND="noninteractive" apt-get install --no-install-recommends -y \
+    build-essential ca-certificates curl gcc git libssl-dev pkg-config ssh \
+    clang llvm nasm \
+    # cleanup
+    && apt-get clean && rm -rf /var/lib/apt/lists/*
+
+# Install rustup and a fixed version of Rust.
```
c3d, line 18: I don't mind, but shouldn't you do a run with multiple steps just to be consistent with the comments on lines 8-10? 😅

c3d (approved):
Thanks @jiangliu . It's so easy to review and approve such smaller commits 😀 I really appreciate that it's quite a bit of extra work on your side, but it makes it so much easier for us.

devtools/dev_container/Dockerfile
jiangliu: The dev container is unstable yet, so keeping them in separate lines helps to use the build cache.
We may combine them into one line once it gets stable:)

jiangliu:
> Thanks @jiangliu . It's so easy to review and approve such smaller commits 😀 I really appreciate that it's quite a bit of extra work on your side, but it makes it so much easier for us.

Never minder, it's also a great chance for me to get familiar with the code too:)

devtools/dev_container/Dockerfile
c3d: OK. Makes sense.

jyao1 (approved):

PR #51, opened 2022-02-02
td-layout: define td-shim on-disk and in-memory layout

jiangliu:
Introduce crate td-layout to define ondisk (file/build-time) and in-memory (runtime) layout for td-shim.
    
Also introduce devtools/td-layout-config to generate layout definitions from a json configuration file.
    

jyao1 (approved):

td-layout/src/build_time.rs
```diff
@@ -0,0 +1,65 @@
+// Copyright (c) 2021 Intel Corporation
+//
+// SPDX-License-Identifier: BSD-2-Clause-Patent
+
```
c3d, line 4: Isn't this file supposed to be generated by `td-layout-config`? 

It is missing the
```
// Auto-generated by `td-layout-config`, do not edit manually.
```

td-layout/src/lib.rs
```diff
@@ -0,0 +1,165 @@
...
+impl RuntimeMemoryLayout {
+    pub fn new(memory_top: u64) -> Self {
+        use crate::runtime::*;
+        let current_base = memory_top & !0xfffff;
+
+        if memory_top < MIN_MEMORY_SIZE {
+            panic!("memory_top 0x{:x} is too small", memory_top);
+        }
+
+        let current_base = current_base - TD_PAYLOAD_EVENT_LOG_SIZE as u64;
```
c3d, line 61: Given that you almost always convert the values to `u64`, why define the constants as `u32` to start with?

c3d: Did a quick experiment, seems to work fine, see https://github.com/c3d/td-shim/commit/6d8aa66e809c98d25ad094f10bf6140dad7970de.

td-layout/src/build_time.rs
jiangliu: The above comment is added after generating runtime/build.rs. Will regenerate them.

td-layout/src/lib.rs
jiangliu: Great, how about opening a new on top of this one?

jiangliu: One reason to use u32 instead of u64 it that td-shim always laid below 4G, and these constants may be converted to u64/usize. For 64 bit platforms, usize is the same as u64 though:)

c3d: > One reason to use u32 instead of u64 it that td-shim always laid below 4G, and these constants may be converted to u64/usize. For 64 bit platforms, usize is the same as u64 though:)

I understand that. My comment is more with respect to the Rust type system. I find it ugly to have `constant as u64` repeated all over the place, this could hide actual issues.

c3d (approved):
Looks good to me.

c3d:
@jiangliu I don't have merge privilege, so can't merge this, but this looks good to me.

PR #58, opened 2022-02-03
test-runner: add crates to run tests for td-shim no_std components

jiangliu:
Some td-shim components are compiled with `#[no_std]` and should be run
on physical/virtual machines. The normal rust test harness does not
support such cases, but it provides mechanisms to extend the stand test
harness. A pair of crates, test-runner-server and test-runner-client,
are implemented to run unit test cases for td-shim no_std components.
These components are used for development only, so hosted uner the
devtools directory.

### DIFF
--- a/devtools/Makefile
+++ b/devtools/Makefile
@@ -1,16 +1,24 @@
 build:
-	cargo build ${BUILD_TYPE_FLAG} -p td-layout-config
+	${CARGO} +${STABLE_TOOLCHAIN} build ${BUILD_TYPE_FLAG} -p td-layout-config
+	${CARGO} +${NIGHTLY_TOOLCHAIN} build ${BUILD_TYPE_FLAG} -p test-runner-client
+	${CARGO} +${STABLE_TOOLCHAIN} build ${BUILD_TYPE_FLAG} -p test-runner-server
 
 check:
-	cargo check ${BUILD_TYPE_FLAG} -p td-layout-config
+	${CARGO} +${STABLE_TOOLCHAIN} check ${BUILD_TYPE_FLAG} -p td-layout-config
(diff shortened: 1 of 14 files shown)


jyao1 (approved):

devtools/Makefile
```diff
@@ -1,11 +1,19 @@
 build:
+	${CARGO} +${NIGHTLY_TOOLCHAIN} build ${BUILD_TYPE_FLAG} -p test-runner-client
+	${CARGO} +${STABLE_TOOLCHAIN} build ${BUILD_TYPE_FLAG} -p test-runner-server
 
 check:
+	${CARGO} +${NIGHTLY_TOOLCHAIN} check ${BUILD_TYPE_FLAG} -p test-runner-client
```
c3d, line 6: The `+` option only works for a `cargo` installed with `rustup`. With a basic Fedora install, that does not work. I'm looking into how to address that.

jiangliu:
```
rustup automatically determines which [toolchain](https://rust-lang.github.io/rustup/concepts/toolchains.html) to use when one of the installed commands like rustc is executed. There are several ways to control and override which toolchain is used:

- A [toolchain override shorthand](https://rust-lang.github.io/rustup/overrides.html#toolchain-override-shorthand) used on the command-line, such as cargo +beta.
- The RUSTUP_TOOLCHAIN environment variable.
- A [directory override](https://rust-lang.github.io/rustup/overrides.html#directory-overrides), set with the rustup override command.
- The [rust-toolchain.toml](https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file) file.
- The [default toolchain](https://rust-lang.github.io/rustup/overrides.html#default-toolchain).
```
How about the second and third methods above?

jiangliu: @c3d anymore comments about this PR? Or could we move to merge it and improve it later?

jyao1:
Merge since it is already 7days.

PR #60, opened 2022-02-03
Refactor tdx-exception into td-exception

jiangliu:
Refactor tdx-exception into td-exception by:
1) rename it to td-exception
2) introduce feature "tdx"
3) add integration test for it


tests/test-td-paging/src/lib.rs
```diff
@@ -55,11 +55,14 @@ mod tests {
     #[test_case]
     fn test_create_paging() {
         td_paging::init();
+        // TODO: make this work
```
jyao1, line 58: Will you continue to make it work in this PR ?

jiangliu: I need help on this:)
I have tried several times but fail, it would be great if you could give some help here:)

jiangliu: I will try to fix the issue for test-pd-paging, but help is needed for the test-td-exception:)

jiangliu: I have tried to fix it, but it need more work help to enable paging.
Help is welcomed so I could move on to other parts:)
How about leaving it as a TODO?

jyao1: agree

PR #65, opened 2022-02-04
Refactor the uefi-pi and r-uefi-pi crates

jiangliu:
Refactor the uefi-pi and r-uefi-pi crates:
1) merge r-uefi-pi into uefi-pi
2) add more test cases
3) improve code readability


jyao1:
Would you please clarify why we need "merge r-uefi-pi into uefi-pi" ?

The original thought is: 
1) `r-uefi-pi` is the definition for hob and fv - that follows UEFI PI specification. We shall NOT change it.
2) `uefi-pi` is the function around hob and fv - no specification defined. We have freedom to add/delete/update them.

jiangliu:
> Would you please clarify why we need "merge r-uefi-pi into uefi-pi" ?
> 
> The original thought is:
> 
> 1. `r-uefi-pi` is the definition for hob and fv - that follows UEFI PI specification. We shall NOT change it.
> 2. `uefi-pi` is the function around hob and fv - no specification defined. We have freedom to add/delete/update them.
> 
> If we merge, I suggest we add some comment for the data structure which follows the UEFI PI specification.

Just hope to reduce the crates number. 
We may isolate the original content from  r-uefi-pi under the namespace of uefi-pi::pi, and all other content directly under uefi-pi. I will give a try.

jyao1:
> Just hope to reduce the crates number.
> We may isolate the original content from r-uefi-pi under the namespace of uefi-pi::pi, and all other content directly under uefi-pi. I will give a try.

OK. Makes sense now. We can reduce crates number.
If we can have a way to isolate "specification defined, which must not be changed" from "our implementation defined, which may be changed", that should be great.

jyao1 (approved):

PR #73, opened 2022-02-05
Move ResetVector to a dedicated crate

jiangliu:
Move ResetVector to a dedicated crate


jyao1:
Any special reason to introduce a new crate?

I have concern because then we need document multiple interfaces, such as BSP->C code, mailbox wakeup etc.

jiangliu:
> Any special reason to introduce a new crate?
> 
> I have concern because then we need document multiple interfaces, such as BSP->C code, mailbox wakeup etc.

The idea is to isolate platform dependent code from platform independent code.
We have three major td modules: bootstrap, shim and payload. 
The payload should be platform independent, logically.
By splitting bootstrap from td-shim, the td-shim may become platform independent too.

So the tdx-bootstrap will host ResetVector and other TDX specific code, and we may use feature to control platform dependent code in td-shim:)

jyao1:
> td-shim may become platform independent too.

Personally, I feel it is an overdesign and I am not sure if it is feasible.
For example, the ParkAp code in reset vector is bound to tdshim mp.rs.

I am OK to rename and refactoring.

But I don't think we need create a dedicated crate - at least not at this moment.

jyao1:
I think we can isolate the code. Or any refactoring/rename is welcome.

My concern is to create a standalone crate, because a standalone crate means a clear interface.

If we really really want to create a standalone crate, I think we must document in td-bootstrap clearly, including but not limited to:
1) MP wakeup API
2) memory usage and readiness
3) stack/heap state
4) TD-SHIM spec dependency

That becomes a mini Td-Shim spec ...

jodh-intel:
Hi @jiangliu - what do we need to do to progress this?

PR #77, opened 2022-02-07
The second part of td-shim refactor

jiangliu:
The second part of td-shim refactor:
1) share common code among td-shim-ld. td-shim-enroll-key, td-shim-sign-payload and rust-tdshim/verifier.
2) refine tcg/verifier.
3) add feature "tdx" to rust-tdshim.


jyao1:
I am OK to merge. But I feel it is challenge work to isolate feature TDX.
There are other features should be under feature TDX, but they are not now...

PR #79, opened 2022-02-09
td-shim: add more kernel image type to KernelInfo HOB

jiangliu:
Add more kernel image type to the KernelInfo HOB, for vmlinux and auto-detection.

### DIFF
--- a/doc/tdshim_spec.md
+++ b/doc/tdshim_spec.md
@@ -363,10 +363,13 @@ typedef enum {
   //   plus 0x200
   PayloadImageTypeBzImage,
 
+  // Payload Binary is vmlinux, follow Linux boot protocol.
+  // It's an ELF64 binary image.
+  PayloadImageTypeVmLinux,
+
   // Payload Binary is VMM loaded vmLinux, follow Linux boot protocol.
   // The entrypoint is defined at HOB_PAYLOAD_INFO_TABLE.Entrypoint.
   PayloadImageTypeRawVmLinux,
-
 
 } PAYLOAD_IMAGE_TYPE;
 


jyao1:
1. Can we adjust the order in enum - move PayloadImageTypeVmLinux before PayloadImageTypeRawVmLinux?

2. I am not sure if we really need PayloadImageTypeAutoDetect.
Can we use "absence of KernelInfoHob" to indicate "auto detect" ?

For example:
If the VMM does not provide the payload image or the VMM does not know the
payload image format, the VMM shall not create this payload info HOB. The TdShim
- must know the payload format and transfer to the payload in right way.
+ must know the payload format or perform auto-detection, then transfer to the
+ payload in right way.

jiangliu:
> 1. Can we adjust the order in enum - move PayloadImageTypeVmLinux before PayloadImageTypeRawVmLinux?
Actually I hope we could reorder these types as:
1) TdshimPayload
2) GenericLinux
3) BzImage
4) Vmlinux
5) RawVmLinux


> 2. I am not sure if we really need PayloadImageTypeAutoDetect.
>    Can we use "absence of KernelInfoHob" to indicate "auto detect" ?
> 
> For example: If the VMM does not provide the payload image or the VMM does not know the payload image format, the VMM shall not create this payload info HOB. The TdShim - must know the payload format and transfer to the payload in right way. + must know the payload format or perform auto-detection, then transfer to the + payload in right way.

I have thought about this, but we definitely need a flag to distinguish TDshim payload and linux kernel because they have different call convention. So it may work with following rules:
1) it's a tdshim payload when the Kernel info HOB is absent
2) Linux kernel always requires the Kernel Info Hob.

jiangliu:
And if desired, an extra entry point may be implemented to convert a linux kernel into a tdshim payload.

jyao1:
Maybe I misunderstand auto-detect. Do you mean "auto-detect == BzImage or Vmlinux"?
Then I am not sure the value to introduce auto-detect.

jiangliu:
> Maybe I misunderstand auto-detect. Do you mean "auto-detect == BzImage or Vmlinux"? Then I am not sure the value to introduce auto-detect.

It's not a bigger deal, a better way is to let td-shim-ld to detect the image type, then we could simplify td-shim implementation.

jyao1:
> * TdshimPayload
> * GenericLinux
> * BzImage
> * Vmlinux
> * RawVmLinux

What is difference between GenericLinux and RawVmLinux ?

jiangliu:
> > * TdshimPayload
> > * GenericLinux
> > * BzImage
> > * Vmlinux
> > * RawVmLinux
> 
> What is difference between GenericLinux and RawVmLinux ?

GenericLinux is either vmlinux or bzImage. RawVmLinux is a pre-preprocessed  raw binary.
I changed my mind. We should leave the complexity to td-shim tools instead of the td-shim firmware. So removed GenericLinux.

jyao1:
My thought is:
* TdshimPayload <== Shim defined boot protocol + PE/ELF format
* GenericLinux <== linux protocol + entrypoint
* BzImage <== linux protocol + BzImage format
* Vmlinux <== linux protocol + Vmlinux format

jiangliu:
> My thought is:
> 
> * TdshimPayload <== Shim defined boot protocol + PE/ELF format
> * GenericLinux <== linux protocol + entrypoint in HOB
> * BzImage <== linux protocol + BzImage format
> * Vmlinux <== linux protocol + Vmlinux format
> 
> Just rename RawVmLinux to GenericLinux.

For GenreicLinux, it also follows linux boot protocol. Not familiar with linux images which accepts HOB list.

jyao1 (approved):

jyao1:
Let's add Vmlinux at first.
The rest may be discussed later.

PR #80, opened 2022-02-10
The fourth part of  td-shim refactor

jiangliu:
1) combine td-shim-ld/td-shim-enroll-key/td-shim-sign-payload into a unified crate td-shim-tools
2) define all td-shim spec defined constants/structure in td-shim.


doc/tdshim_spec.md
```diff
@@ -267,6 +267,14 @@ typedef enum {
...
   PayloadImageTypeRawVmLinux,
 
+  // Payload Binary is vmlinux, follow Linux boot protocol.
+  // It's an ELF64 binary image.
+  PayloadImageTypeVmLinux,
+
+  // Auto-detect type of the payload image, follow Linux boot protocol.
+  // It's one of bzImage, vmLinux. RawVmLinux should be explicitly marked
+  // instead of auto-detection.
+  PayloadImageTypeAutoDetect,
```
jyao1, line 277: Should it be align with latest spec in main?

jiangliu: Let first merge the spec, then update the code:)

jyao1: i mean the spec should be chanaged - PayloadImageTypeAutoDetect should be removed

jiangliu: Sorry, the last patch should be removed:) 
Fixed.

PR #83, opened 2022-02-11
[main] td-uefi-pi: a simple wrapper over r-efi for td-shim

jiangliu:
The td-uefi-pi is a simple wrapper crate over r-efi, which provides:
1) constants and structures defined by the uefi specification
2) helper methods to handle uefi firmware volumes
3) helper methods to handle uefi HOB list.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,6 +8,7 @@ members = [
 	"devtools/test-runner-server",
 	"td-layout",
 	"td-paging",
+	"td-uefi-pi",
 	"tdx-tdcall",
 ]
 
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ else
 	export BUILD_TYPE_FLAG=
 endif
 
-GENERIC_LIB_CRATES = td-layout
+GENERIC_LIB_CRATES = td-layout td-uefi-pi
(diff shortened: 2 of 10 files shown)


jyao1:
Just some clarification:
FV and HOB are UEFI-PI concept, not UEFI concept.
Actually, you are creating wrap for UEFI-PI, not UEFI.

The difference is: UEFI is an interface spec between BIOS/Firmware and OS.
UEFI-PI is an interface spec inside of BIOS/Firmware.

Would you mind to rename to td-uefipi ?

jiangliu:
> Just some clarification: FV and HOB are UEFI-PI concept, not UEFI concept. Actually, you are creating wrap for UEFI-PI, not UEFI.
> 
> The difference is: UEFI is an interface spec between BIOS/Firmware and OS. UEFI-PI is an interface spec inside of BIOS/Firmware.
> 
> Would you mind to rename to td-uefipi ?

sorry, I'm not familiar with uefi spec:)
Will rename it.
What I want is to have a common prefix for crates, it may help us to publish these crates to crates.io later.

jyao1:
yes, agree the direction. Just naming concern.

UEFI people may get surprised to find a UEFI crate, but all thing there is PI stuff.

jyao1 (approved):

Issue #84, opened 2022-02-11
Combine elf-loader/pe-loader into td-loader


jiangliu:
Combine elf-loader/pe-loader into td-loader, so we have little crates to manage.

PR #89, opened 2022-02-14
td-shim: Combine elf-loader/pe-loader into td-loader

liuw1:
Combine elf-loader/pe-loader into td-loader,
so we have little crates to manage.

Fix: #84


td-shim/Cargo.toml
```diff
@@ -49,9 +48,9 @@ cet-ss = []
 secure-boot = ["der"]
 tdx = ["tdx-tdcall", "td-exception/tdx", "td-logger/tdx", "x86"]
 main = [
-    "elf-loader",
+    "td-loader",
```
jiangliu, line 51: only one copy of "td-loader" is needed, please remove the duplicated one.

liuw1: Thanks for your comment. I have update the PR.

jiangliu (approved):

PR #90, opened 2022-02-14
td-shim: set HARDWARE_ENABLED_EXECUTION_DISABLE to be true

gaojiaqi7:
Since EFER MSR cannot be set for now by TD, td-shim will break when
running into enable_execution_disable_bit().

In TDX, NX is enabled by default, so HARDWARE_ENABLED_EXECUTION_DISABLE
should be set to 'true'.


jyao1:
enable_execute_disable_bit() is wrong. the NX is always enabled by TDX module. Not need to set it.

Please remove the MSR access. That code will cause exception.

jiangliu (approved):

PR #91, opened 2022-02-14
Keep same as before refactoring.

### LINKED ISSUE #92: regression: MSR access cause exception
MSR access will cause VE exception that cause td-shim failure. 

https://github.com/confidential-containers/td-shim/blob/staging/td-shim/src/bin/td-shim/td/tdx.rs

```
fn enable_execute_disable_bit() {
    let mut msr: u64;

    unsafe { msr = asm_read_msr64(0xC0000080) };
    msr |= 0x800;
    unsafe { asm_write_msr64(0xC0000080, msr) };
}
```

According to TDX spec, 
1) the MSR access should be done by VMCALL, instead of MSR instruction.
2) the TDX module will enable NX by default. Not need to use MSR to enable them again.

This code is redundant and shall be removed.

### LINKED ISSUE #93: regression: HLT causes double exception
See code below in exception handle.

https://github.com/confidential-containers/td-shim/blob/staging/td-exception/src/interrupt.rs
```
fn deadloop() {
    // TBD: empty `loop {}` wastes CPU cycles
    #[allow(clippy::empty_loop)]
    loop {
        x86_64::instructions::interrupts::enable();
        x86_64::instructions::hlt();
    }
}
```

According to TDX spec, the HLT access should be done by VMCALL, instead of HLT instruction.

Using HLT directly will cause reentry to exception handle.

xiaoyuxlu:
Fix #92 
Fix #93

Fix regression test failure.

### DIFF
--- a/td-exception/src/interrupt.rs
+++ b/td-exception/src/interrupt.rs
@@ -406,7 +406,9 @@ fn deadloop() {
     // TBD: empty `loop {}` wastes CPU cycles
     #[allow(clippy::empty_loop)]
     loop {
-        x86_64::instructions::interrupts::enable();
-        x86_64::instructions::hlt();
+        // // Keep the same as before refactoring.
+        // x86_64::instructions::interrupts::enable();
+        // x86_64::instructions::hlt();
+        //
     }
 }
--- a/td-shim/src/bin/td-shim/td/tdx.rs
+++ b/td-shim/src/bin/td-shim/td/tdx.rs
@@ -11,39 +11,8 @@ extern "win64" {
 
 const EXTENDED_FUNCTION_INFO: u32 = 0x80000000;
 const EXTENDED_PROCESSOR_INFO: u32 = 0x80000001;
-const HARDWARE_ENABLED_EXECUTION_DISABLE: bool = false;
+
 const SHA384_DIGEST_SIZE: usize = 48;
-
-fn is_execute_disable_bit_available() -> bool {
-    let cpuid = unsafe { core::arch::x86_64::__cpuid(EXTENDED_FUNCTION_INFO) };
-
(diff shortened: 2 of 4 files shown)


jyao1:
enable_execute_disable_bit() is wrong. the NX is always enabled by TDX module. Not need to set it.

Please remove the MSR access. That code will cause exception.

jyao1:
@jiangliu, could you please take a look for this ?

td-exception/src/interrupt.rs
```diff
@@ -406,7 +406,9 @@ fn deadloop() {
     // TBD: empty `loop {}` wastes CPU cycles
     #[allow(clippy::empty_loop)]
     loop {
```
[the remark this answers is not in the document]
jyao1: See #93.

A TD cannot run HLT instruction directly.

td-shim/src/bin/td-shim/td/tdx.rs
```diff
@@ -11,39 +11,8 @@ extern "win64" {
 
 const EXTENDED_FUNCTION_INFO: u32 = 0x80000000;
 const EXTENDED_PROCESSOR_INFO: u32 = 0x80000001;
-const HARDWARE_ENABLED_EXECUTION_DISABLE: bool = false;
```
jiangliu, line 14: The issue could also get fixed by changing "HARDWARE_ENABLED_EXECUTION_DISABLE" to true, as in #90 :)

It would be great to document that `excution_disable` is actually enabled by hardware.

jyao1: Having excution_disable() API is misleading. I don't think we need this function.
Yes, we should document it.

jiangliu (approved):

Issue #95, opened 2022-02-14
regression: build failed, error[E0433]: failed to resolve: use of undeclared crate or module `td_layout`


xiaoyuxlu:
Follow the compilation steps in the readme  https://github.com/confidential-containers/td-shim/blob/3a3bfe5fbed2a1c942283fd9a8fcff2be6969514/README.md#build-tdshim will result in the following error:

```
error[E0433]: failed to resolve: use of undeclared crate or module `td_layout`
  --> td-shim/src/reset_vector.rs:10:5
   |
10 | use td_layout::build_time::TD_SHIM_RESET_VECTOR_SIZE;
   |     ^^^^^^^^^ use of undeclared crate or module `td_layout`

error[E0425]: cannot find value `TD_SHIM_RESET_VECTOR_SIZE` in this scope
  --> td-shim/src/reset_vector.rs:47:13
   |
47 |             TD_SHIM_RESET_VECTOR_SIZE + size_of::<ResetVectorHeader>() as u32,
   |             ^^^^^^^^^^^^^^^^^^^^^^^^^ not found in this scope

error[E0425]: cannot find value `TD_SHIM_RESET_VECTOR_SIZE` in this scope
  --> td-shim/src/reset_vector.rs:58:13
   |
58 |             TD_SHIM_RESET_VECTOR_SIZE + size_of::<CommonSectionHeader>() as u32,
   |             ^^^^^^^^^^^^^^^^^^^^^^^^^ not found in this scope

Some errors have detailed explanations: E0425, E0433.
For more information about an error, try `rustc --explain E0425`.
error: could not compile `td-shim` due to 3 previous errors
```

jiangliu:
Strange, this should have been detected by the CI system. Something wrong with the CI?

PR #96, opened 2022-02-14
td-shim: fix a dependency issue about td-layout

jiangliu:
Fix a compilation failure caused by incorrect dependency.

Fixes: #95

### DIFF
--- a/td-shim-tools/Cargo.toml
+++ b/td-shim-tools/Cargo.toml
@@ -35,7 +35,7 @@ ring = { git = "https://github.com/jyao1/ring.git", branch = "uefi_support", def
 
 [features]
 default = ["boot-kernel", "enroller", "linker", "signer"]
-boot-kernel = ["td-layout/boot-kernel"]
+boot-kernel = ["td-layout/boot-kernel", "td-shim/boot-kernel"]
 enroller = ["clap", "der", "env_logger", "log", "ring"]
 linker = ["clap", "env_logger", "log", "pe-loader"]
 signer = ["clap", "der", "env_logger", "log", "ring"]
--- a/td-shim/Cargo.toml
+++ b/td-shim/Cargo.toml
@@ -20,6 +20,7 @@ lazy_static = { version = "1.4.0", features = ["spin_no_std"] }
 log = "0.4.13"
 r-efi = "3.2.0"
 scroll = { version = "0.10", default-features = false, features = ["derive"] }
+td-layout = { path = "../td-layout" }
 uefi-pi =  { path = "../uefi-pi" }
 zerocopy = "0.6.0"
 
@@ -29,7 +30,6 @@ ring = { git = "https://github.com/jyao1/ring.git", branch = "uefi_support", def
 pe-loader = { path = "../pe-loader", optional = true }
 spin = { version = "0.9.2", optional = true }
 td-exception =  { path = "../td-exception", features = ["tdx"], optional = true }
-td-layout = { path = "../td-layout", optional = true }
 td-logger =  { path = "../td-logger", optional = true }
 td-paging = { path = "../td-paging", optional = true }
 x86 = { version ="0.44.0", optional = true }
@@ -55,7 +55,6 @@ main = [
     "ring",
     "spin",
     "td-exception",
-    "td-layout",
     "td-logger",
     "td-paging",
     "x86",


xiaoyuxlu:
Two other ways are available:

1.  Provide a unified features at compile time. 

```
TD_SHIM_FEATURES=" --features=main,tdx"
TD_SHIM_TOOL_FEATURES=" --features=td-shim/main,td-shim/tdx,td-shim/default"
cargo build --verbose -Zbuild-std=core,compiler_builtins,alloc -Zbuild-std-features=compiler-builtins-mem --target x86_64-unknown-uefi --release -p ${TD_SHIM} ${TD_SHIM_FEATURES}
cargo build --verbose -Zbuild-std=core,alloc,compiler_builtins -Zbuild-std-features=compiler-builtins-mem --target x86_64-unknown-uefi --release -p ${PAYLOAD_NAME} ${TD_SHIM_FEATURES}
cargo run --verbose -p td-shim-tools --bin td-shim-ld ${TD_SHIM_TOOL_FEATURES} -- target/x86_64-unknown-uefi/release/ResetVector.bin target/x86_64-unknown-uefi/release/${TD_SHIM}.efi $PAYLOAD_BIN -o target/x86_64-unknown-uefi/release/final.bin
```

2. add `td-shim/default features` to td-shim-tool `td-shim-tools/Cargo.toml`:

```
[features]
default = ["boot-kernel", "enroller", "linker", "signer", "td-shim/default"]
```

Which do you think is the best?

jiangliu:
> cargo run --verbose -p td-shim-tools --bin td-shim-ld ${TD_SHIM_TOOL_FEATURES} -- target/x86_64-unknown-uefi/release/ResetVector.bin target/x86_64-unknown-uefi/release/${TD_SHIM}.efi $PAYLOAD_BIN -o target/x86_64-unknown-uefi/release/final.bin

I followed you steps, but without the "TD_SHIM_TOOL_FEATURES" flag set.
```
cargo build --verbose -Zbuild-std=core,compiler_builtins,alloc -Zbuild-std-features=compiler-builtins-mem --target x86_64-unknown-uefi --release -p td-shim --features=tdx,main
cargo build --verbose -Zbuild-std=core,compiler_builtins,alloc -Zbuild-std-features=compiler-builtins-mem --target x86_64-unknown-uefi --release -p td-payload --features=tdx,main
cargo run --verbose -p td-shim-tools --bin td-shim-ld -- target/x86_64-unknown-uefi/release/ResetVector.bin target/x86_64-unknown-uefi/release/td-shim.efi target/x86_64-unknown-uefi/release/td-payload.efi -o target/x86_64-unknown-uefi/release/final.bin
```

So still not sure about why we need the "TD_SHIM_TOOL_FEATURES".

xiaoyuxlu:
Because td-shim-tool depends on td-shim. If the features of these two components are inconsistent, it may cause problems in generating the image. 
It is assumed that the structure size of the td-shim varies according to the feature. td-shim uses features A but td-shim-tool does not use features A. In this way, the size of a structure in td-shim and td-shim-tool will be inconsistent. It may cause problems with the image generated by td-shim-tools. Am I expressing that clearly?

jiangliu:
> Because td-shim-tool depends on td-shim. If the features of these two components are inconsistent, it may cause problems in generating the image. It is assumed that the structure size of the td-shim varies according to the feature. td-shim uses features A but td-shim-tool does not use features A. In this way, the size of a structure in td-shim and td-shim-tool will be inconsistent. It may cause problems with the image generated by td-shim-tools. Am I expressing that clearly?

Good point! It must keep consistence between td-shim and td-shim-tools.

There are several design goals with the td-shim crate:
1) td-shim is both a library and a binary. The `main` feature is used to control dependencies for the binary.
2) the td-shim library follows the td-shim spec, and should define all external visible behaviors.
3) the td-shim library should compatible with stable rust toolchains, so we could build td-rust-tools with stable tool chains.

With above design goals,
1) we should not enable td-shim/main feature for td-shim-tools
2) we may keep the default features for td-shim/td-shim-tools in sync, so it works with the default configuration.
3) end user takes the responsibility to ensure consistent feature flags for td-shim/td-shim-tools.

xiaoyuxlu:
> > Because td-shim-tool depends on td-shim. If the features of these two components are inconsistent, it may cause problems in generating the image. It is assumed that the structure size of the td-shim varies according to the feature. td-shim uses features A but td-shim-tool does not use features A. In this way, the size of a structure in td-shim and td-shim-tool will be inconsistent. It may cause problems with the image generated by td-shim-tools. Am I expressing that clearly?
> 
> Good point! It must keep consistence between td-shim and td-shim-tools.
> 
> There are several design goals with the td-shim crate:
> 
>     1. td-shim is both a library and a binary. The `main` feature is used to control dependencies for the binary.
> 
>     2. the td-shim library follows the td-shim spec, and should define all external visible behaviors.
> 
>     3. the td-shim library should compatible with stable rust toolchains, so we could build td-rust-tools with stable tool chains.
> 
> 
> With above design goals,
> 
>     1. we should not enable td-shim/main feature for td-shim-tools
> 
>     2. we may keep the default features for td-shim/td-shim-tools in sync, so it works with the default configuration.
> 
>     3. end user takes the responsibility to ensure consistent feature flags for td-shim/td-shim-tools.

That's clear, thanks.

Reviewed-by: Xiaoyu Lu <xiaoyu1.lu@intel.com>

PR #101, opened 2022-02-17
[main] td-logger: a simple logger backend for td-shim

jiangliu:
A simple logger backend for td-shim, based on serial port.

### DIFF
--- a/td-logger/src/lib.rs
+++ b/td-logger/src/lib.rs
@@ -1,0 +1,91 @@
+// Copyright (c) 2020 Intel Corporation
+//
+// SPDX-License-Identifier: BSD-2-Clause-Patent
+
+#![no_std]
+
+use log::{Level, Metadata, Record};
+use log::{LevelFilter, SetLoggerError};
+
+mod logger;
+pub use logger::*;
+
+macro_rules! tdlog {
+    ($($arg:tt)*) => (crate::logger::_log_ex(crate::logger::LOG_LEVEL_INFO, crate::logger::LOG_MASK_ALL, format_args!($($arg)*)));
+}
+
+/// Logger backend for td-shim.
+pub struct LoggerBackend;
+
+impl log::Log for LoggerBackend {
+    fn enabled(&self, metadata: &Metadata) -> bool {
+        metadata.level() <= Level::Info
+    }
+
+    fn log(&self, record: &Record) {
+        if self.enabled(record.metadata()) {
+            tdlog!("{} - {}", record.level(), record.args());
+        }
+    }
+
+    fn flush(&self) {}
+}
+
+/// Logger backend for the log crate
+static LOGGER_BACKEND: LoggerBackend = LoggerBackend;
+
+pub fn init() -> Result<(), SetLoggerError> {
+    dbg_write_string("logger init\n");
+    log::set_logger(&LOGGER_BACKEND).map(|()| log::set_max_level(LevelFilter::Info))
+}
+
+/// Write a byte to the debug port, converting `\n' to '\r\n`.
+pub fn dbg_write_byte(byte: u8) {
+    if byte == b'\n' {
+        dbg_port_write(b'\r')
+    }
+    dbg_port_write(byte)
+}
+
+/// Write a string to the debug port.
+pub fn dbg_write_string(s: &str) {
+    for c in s.chars() {
+        dbg_write_byte(c as u8);
+    }
+}
+
+#[cfg(any(feature = "tdx", feature = "serial-port"))]
+const SERIAL_IO_PORT: u16 = 0x3F8;
+
+#[cfg(feature = "tdx")]
+fn dbg_port_write(byte: u8) {
+    tdx_tdcall::tdx::tdvmcall_io_write_8(SERIAL_IO_PORT, byte);
+}
+
+#[cfg(all(not(feature = "tdx"), feature = "serial-port"))]
+fn dbg_port_write(byte: u8) {
+    unsafe { x86::io::outb(SERIAL_IO_PORT, byte) };
+}
+
+#[cfg(all(not(feature = "tdx"), not(feature = "serial-port")))]
+fn dbg_port_write(_byte: u8) {}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_logger() {
+        init().unwrap();
+
+        assert_eq!(LOGGER.lock().get_level(), LOG_LEVEL_INFO);
+        LOGGER.lock().set_level(LOG_LEVEL_ERROR);
+        assert_eq!(LOGGER.lock().get_level(), LOG_LEVEL_ERROR);
+
+        assert_eq!(LOGGER.lock().get_mask(), LOG_MASK_ALL);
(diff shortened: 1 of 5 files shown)


td-logger/src/logger.rs
```diff
@@ -0,0 +1,95 @@
...
+        self.level
+    }
+    pub fn set_level(&mut self, level: usize) {
+        self.level = level;
+    }
+
+    pub fn get_mask(&self) -> u64 {
+        self.mask
+    }
+    pub fn set_mask(&mut self, mask: u64) {
```
jodh-intel, line 68: Nit: Missing blank line.

td-logger/src/logger.rs
```diff
@@ -0,0 +1,95 @@
...
+        Ok(())
+    }
+}
+
+/// Log the message with level and subsystem filtering.
+pub fn _log_ex(level: usize, mask: u64, args: fmt::Arguments) {
+    if level > LOGGER.lock().get_level() {
+        return;
+    }
+    if (mask & LOGGER.lock().get_mask()) == 0 {
```
jodh-intel, line 85: Currently every log call requires the lock to be obtained twice. I think you can simplify this to:

```suggestion
    let logger = LOGGER.lock();
    
    if level > logger.get_level() {
        return;
    }
    
    if (mask & logger.get_mask()) == 0 {
```

td-logger/src/logger.rs
```diff
@@ -0,0 +1,95 @@
...
+        return;
+    }
+    _log(args);
+}
+
+/// Log the message without level and subsystem filtering.
+pub fn _log(args: fmt::Arguments) {
+    use core::fmt::Write;
+    LOGGER.lock().write_fmt(args).unwrap();
```
jodh-intel, line 94: Oh - you actually have to get the lock *three times* for every log call!

td-logger/src/logger.rs
```diff
@@ -0,0 +1,95 @@
...
+
+use crate::{dbg_port_write, dbg_write_string};
+
+pub const LOG_LEVEL_VERBOSE: usize = 1000;
+pub const LOG_LEVEL_INFO: usize = 100;
+pub const LOG_LEVEL_WARN: usize = 10;
+pub const LOG_LEVEL_ERROR: usize = 1;
+pub const LOG_LEVEL_NONE: usize = 0;
+
+pub const LOG_MASK_COMMON: u64 = 0x1;
```
jodh-intel, line 19: What does this actually log?

jodh-intel (commented):
Thanks @jiangliu.

td-logger/src/logger.rs
```diff
@@ -0,0 +1,95 @@
...
+// Use lazy_static here to prevent potential problems caused by compiler/linker optimization.
+lazy_static! {
+    pub static ref LOGGER: Mutex<Logger> = Mutex::new(Logger {
+        level: LOG_LEVEL_VERBOSE,
+        mask: LOG_MASK_ALL,
+    });
+}
+
+pub struct Logger {
+    level: usize,
```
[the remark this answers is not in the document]
jiangliu: It follow the style of the log crate:)

td-logger/src/logger.rs
jiangliu: Fixed

td-logger/src/logger.rs
jiangliu: fixed

td-logger/src/logger.rs
jiangliu: Something "generic", or "other", maybe:)
Depend on the caller.

td-logger/src/logger.rs
```diff
@@ -0,0 +1,93 @@
...
+pub const LOG_MASK_EVENT: u64 = 0x8;
+pub const LOG_MASK_IMAGE: u64 = 0x10;
+// Core - Runtime Service (BIT16 ~ BIT 23)
+pub const LOG_MASK_VARIABLE: u64 = 0x10000;
+// Core - Console (BIT24 ~ BIT 31)
+pub const LOG_MASK_CONOUT: u64 = 0x1000000;
+pub const LOG_MASK_CONIN: u64 = 0x2000000;
+// Protocol - (BIT32 ~ BIT63)
+pub const LOG_MASK_BLOCK_IO: u64 = 0x100000000;
+pub const LOG_MASK_FILE_SYSTEM: u64 = 0x200000000;
```
jyao1, line 32: Above (from LOG_MASK_PROTOCOL to LOG_MASK_FILE_SYSTEM) seems UEFI concept.

I am not sure they are absolutely needed.

To simplify the design, how about we can remove them at this moment ?
If there is need, we can add based upon real use case.

jiangliu: Sounds reasonable.

jyao1 (approved):

td-logger/src/logger.rs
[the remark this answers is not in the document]
jodh-intel: I'm still not clear on this. The `log` crate uses an `enum` for level and that (in this case) would resolve to a `u8` I think?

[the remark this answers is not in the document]
jiangliu: The answer may be "we don't really care about the size of the field":)
It's a singleton, and the `Logger` may always occupies 16-bytes due to alignment:)

td-logger/src/lib.rs
```diff
@@ -0,0 +1,91 @@
...
+#![no_std]
+
+use log::{Level, Metadata, Record};
+use log::{LevelFilter, SetLoggerError};
+
+mod logger;
+pub use logger::*;
+
+macro_rules! tdlog {
+    ($($arg:tt)*) => (crate::logger::_log_ex(crate::logger::LOG_LEVEL_VERBOSE, crate::logger::LOG_MASK_COMMON, format_args!($($arg)*)));
```
jodh-intel, line 14: Question: Just checking that the first two params are as expected. It's fairly conventional to log at info level by default but we're logging at verbose level. So should verbose => info, or should `LOG_MASK_COMMON` => `LOG_MASK_ALL`?

td-logger/src/logger.rs
```diff
@@ -0,0 +1,79 @@
...
+
+//! A standalone logger to directly log messages, independent of the log crate.
+
+use core::fmt::{self, Write};
+use lazy_static::lazy_static;
+use spin::Mutex;
+
+use crate::{dbg_port_write, dbg_write_string};
+
+pub const LOG_LEVEL_VERBOSE: usize = 4;
```
jodh-intel, line 13: Nit: since we have a lot of log level bits to play with, should this be split into `LOG_LEVEL_DEBUG` and `LOG_LEVEL_TRACE`? That would align with the log crates levels:

- https://docs.rs/log/0.4.14/log/enum.Level.html

jodh-intel (approved):
Thanks @jiangliu. A few nits and a question but overall...

lgtm

jyao1:
Any further discussion or update?

td-logger/src/logger.rs
jiangliu: fixed

Issue #110, opened 2022-02-21
build script should check dependencies and fail early


jodh-intel:
Building the shim is currently somewhat awkward since you need:

- clang ([GCC won't work](https://github.com/confidential-containers/td-shim/issues/109)).
- nasm
- `cargo xbuild`

Admittedly, this _is_ documented in https://github.com/confidential-containers/td-shim/blob/staging/README.md, but we should make it easier for users by checking for the expected commands in `build.rs` so that the build can fail early (rather than letting it chug on and die in flames at some later stage).

Related:

- https://github.com/confidential-containers/td-shim/pull/2#issuecomment-977991422
- https://github.com/confidential-containers/td-shim/pull/2#discussion_r759137136

jyao1:
@jodh-intel, can you take this and propose a patch?

jodh-intel:
Yes, I can look at this, but I don't have privs to assign this issue to myself.

@sameo - could you possibly assign to me or grant me those privs?

PR #116, opened 2022-02-22
td-shim: Check build tooling

jodh-intel:
Although the required tooling is documented in the README, it's too easy
to end up with a failed build due to missing tooling.

Add a new function to the build script to check for all dependencies and
fail with helpful messages on error.

Fixes: #110.

Signed-off-by: James O. D. Hunt <james.o.hunt@intel.com>

### DIFF
--- a/.github/workflows/library.yml
+++ b/.github/workflows/library.yml
@@ -10,6 +10,9 @@ on:
 name: Library Crates
 
 env:
+  AR: llvm-ar
+  AS: nasm
+  CC: clang
   RUST_TOOLCHAIN: nightly-2021-08-20
   TOOLCHAIN_PROFILE: minimal
 
@@ -25,14 +28,15 @@ jobs:
           - ubuntu-20.04
           - windows-2019
     steps:
-      - name: install NASM
-        uses: ilammy/setup-nasm@v1
-
+      # Install first since it's needed to build NASM
       - name: Install LLVM and Clang
         uses: KyleMayes/install-llvm-action@v1
         with:
           version: "10.0"
           directory: ${{ runner.temp }}/llvm
+
+      - name: install NASM
+        uses: ilammy/setup-nasm@v1
 
       - name: Checkout sources
         uses: actions/checkout@v2
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -2,6 +2,11 @@ on: [push, pull_request]
 
 name: Format and Clippy
 
+env:
+  AR: llvm-ar
+  AS: nasm
+  CC: clang
+
 jobs:
   clippy:
     name: Clippy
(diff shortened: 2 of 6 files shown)


jodh-intel:
Examples of error messages for various scenarios:

- Missing assembler
  ```
  ERROR: cannot find assembler command "nasm": CannotFindBinaryPath (expected "nasm")
  ```
- Missing compiler
  ```
  ERROR: compiler command "cc" is not "clang" - do you need to set $CC?
  ```
- Invalid `CC` value
  ```
  ERROR: cannot find compiler command "clang": CannotFindBinaryPath (expected "clang")
  ```
- `CC` not set
  ```
  ERROR: compiler command "cc" is not "clang" - do you need to set $CC?
  ```
- Missing archiver
  ```
  ERROR: cannot find archiver command "llvm-ar": CannotFindBinaryPath (expected "llvm-ar")
  ```  
 


/cc @jyao1.

jodh-intel:
Hi @jyao1 - I've had to change some of the GH actions as they were failing as `CC` wasn't set. The problem is that rust doesn't provide a way for a build script to determine the name of the `cargo` command specified (`check`, `run`, `build`, `clippy`, _etc_), so the new checks added on this PR run always.

However, there is one final GH action failure I can't resolve - do you have any thoughts on the Win 2019 "final.bin" GHA failure?

.github/workflows/library.yml
```diff
@@ -10,6 +10,9 @@ on:
 name: Library Crates
 
 env:
+  AR: llvm-ar
```
jiangliu, line 13: The library action should be run on host instead of inside vm. So seem it doesn't rely on llvm-ar/nasm/clang.
What I hope to achieve is:
1) all components to run on host should be compiled with standard stable toolchain.
2) components which may be run inside the TD env should be compiled with special nightly toolchain/llvm/nasm.

jodh-intel: > The library action should be run on host instead of inside vm. So seem it doesn't rely on llvm-ar/nasm/clang.

@jiangliu - tbh I'm not clear how the GHA scripts were running previously if CC wasn't set for all steps that run `cargo`.

Note that previously, I found this warning in the CI logs (which is no longer shown):

```
warning: Linking two modules of different data layouts: '' is 'e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128' whereas 'td_payload.52f1f334-cgu.9' is 'e-m:e-i64:64-f80:128-n8:16:32:64-S128'
##[warning]Linking two modules of different data layouts: '' is 'e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128' whereas 'td_payload.52f1f334-cgu.9' is 'e-m:e-i64:64-f80:128-n8:16:32:64-S128'
```

I also still cannot work out why the 2019 build is failing. What I could do is add an env check to the build script so that if a var like `TDSHIM_DISABLE_BUILD_CHECKS=1`, the `check_environment()` function isn't called?

Issue #82, opened 2022-02-10
Relocate the MAILBOX page to be continuous to the ACPI/event log regions


jiangliu:
The MAILBOX page should be relocated to upper address to be continuous to the ACPI/event log regions, so the lower region could be released.

gaojiaqi7:
Hi @jiangliu , mailbox is located at 0xFF040000 (can be found in build time layout), do you think it is necessary to relocate it to a memory region adjacent to ACPI/event log?

jiangliu:
> Hi @jiangliu , mailbox is located at 0xFF040000 (can be found in build time layout), do you think it is necessary to relocate it to a memory region adjacent to ACPI/event log?

Yes，we could reuse the memory below mailbox/acpi/event log regions once the relocation is done.

jyao1:
Typically, the ideal memory layout is:
+---------------+
| Usable            |
+---------------+
| ACPI Reclaim |
+---------------+
| ACPI NVS       |
+---------------+
| Reserved        |
+---------------+

PR #120, opened 2022-02-23
Relocate mailbox and AP function

gaojiaqi7:
This PR addresses these issues:
    - Relocate the MAILBOX page to be continuous to the ACPI/event log regions #82
    - Relocate the AP loop function to be continuous to the above area
    - Use APIC ID instead of CPU index to identify AP 

APs are running in the loop waiting for BSP's instruction after they switch to long mode. At before, We assumed that CPU index get from TDCALL is equal to the APIC ID, but these two value may be different in some cases, such as configuring multiple sockets in QEMU.

One solution is to relocate APs' running loop into a new place where they use APIC ID to identify themselves. APs will use TDVMCALL to know their APIC ID and wait for the instructions from BSP to wakeup.

APs will run through two stages:
    - At first stage, APs are waiting for BSP to assign works using CPU index since BSP does not know the APIC ID of APs.
    - After this, BSP will copy a new function into Mailbox memory then wakeup all the APs and instruct them to jump into the new loop where APs will wait for command from OS and use APIC ID to identify themselves with a new mailbox address.

### DIFF
--- a/td-layout/src/memslice.rs
+++ b/td-layout/src/memslice.rs
@@ -8,7 +8,7 @@ use crate::build_time::{
 };
 use crate::runtime::{
     TD_PAYLOAD_ACPI_SIZE, TD_PAYLOAD_BASE, TD_PAYLOAD_EVENT_LOG_SIZE, TD_PAYLOAD_HOB_SIZE,
-    TD_PAYLOAD_SIZE,
+    TD_PAYLOAD_MAILBOX_SIZE, TD_PAYLOAD_SIZE,
 };
 
 /// Type of build time and runtime memory regions.
@@ -25,6 +25,8 @@ pub enum SliceType {
     Payload,
     /// The `TD_HOB` region in runtime memory layout
     PayloadHob,
+    /// The 'Mailbox' region in runtime memory layout
+    RelocatedMailbox,
     /// The `TD_EVENT_LOG` region in runtime memory layout
     EventLog,
     /// The `ACPI` region in runtime memory layout
@@ -95,10 +97,15 @@ pub unsafe fn get_dynamic_mem_slice_mut<'a>(t: SliceType, base_address: usize) -
             base_address as *const u8 as *mut u8,
             TD_PAYLOAD_EVENT_LOG_SIZE as usize,
         ),
+        SliceType::RelocatedMailbox => core::slice::from_raw_parts_mut(
+            base_address as *const u8 as *mut u8,
+            TD_PAYLOAD_MAILBOX_SIZE as usize,
+        ),
         SliceType::Acpi => core::slice::from_raw_parts_mut(
             base_address as *const u8 as *mut u8,
             TD_PAYLOAD_ACPI_SIZE as usize,
         ),
+
         _ => panic!("get_dynamic_mem_slice_mut: not support"),
     }
 }
--- a/td-layout/src/runtime.rs
+++ b/td-layout/src/runtime.rs
@@ -19,18 +19,20 @@
                     |    PAYLOAD   |    (0x02000000)
                     +--------------+
                     |   ........   |
-                    +--------------+ <-  0x7D000000
+                    +--------------+ <-  0x7CFFE000
                     |     DMA      |    (0x01000000)
-                    +--------------+ <-  0x7E000000
+                    +--------------+ <-  0x7DFFE000
                     |     HEAP     |    (0x01000000)
-                    +--------------+ <-  0x7F000000
+                    +--------------+ <-  0x7EFFE000
                     |     STACK    |    (0x00800000)
-                    +--------------+ <-  0x7F800000
+                    +--------------+ <-  0x7F7FE000
                     |      SS      |    (0x00200000)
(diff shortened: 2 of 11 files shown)


td-shim/src/bin/td-shim/asm/ap_loop.asm
```diff
@@ -0,0 +1,90 @@
...
+    mov         r13, 0
+    # TDVMCALL
+    .byte       0x66, 0x0f, 0x01, 0xcc
+    test        rax, rax
+    jnz         panic
+    #
+    # r8 will hold the APIC ID of current AP
+    mov         r8, r15
+
+check_apicid:
```
jodh-intel, line 61: Is it worth making these labels local?

```diff
diff --git a/td-shim/src/bin/td-shim/asm/ap_loop.asm b/td-shim/src/bin/td-shim/asm/ap_loop.asm
index f6e792b..95fce3a 100644
--- a/td-shim/src/bin/td-shim/asm/ap_loop.asm
+++ b/td-shim/src/bin/td-shim/asm/ap_loop.asm
@@ -53,30 +53,30 @@ ap_relocated_func:
     # TDVMCALL
     .byte       0x66, 0x0f, 0x01, 0xcc
     test        rax, rax
-    jnz         panic
+    jnz         .panic
     #
     # r8 will hold the APIC ID of current AP
     mov         r8, r15
 
-check_apicid:
+.check_apicid:
     #
     # Determine if this is a broadcast or directly for my apic-id, if not, ignore
     cmp     dword ptr[rbx + ApicIdOffset], MailboxApicIdBroadcast
-    je      check_command
+    je      .check_command
     cmp     dword ptr[rbx + ApicIdOffset], r8d
-    jne     check_apicid
+    jne     .check_apicid
 
-check_command:
+.check_command:
     mov     eax, dword ptr[rbx + CommandOffset]
     cmp     eax, MpProtectedModeWakeupCommandNoop
-    je      check_apicid
+    je      .check_apicid
 
     cmp     eax, MpProtectedModeWakeupCommandWakeup
-    je      wakeup
+    je      .wakeup
 
-    jmp     check_apicid
+    jmp     .check_apicid
 
-wakeup:
+.wakeup:
     #
     # BSP sets these variables before unblocking APs
     mov     rax, 0
@@ -84,7 +84,7 @@ wakeup:
     nop
     jmp     rax
 
-panic:
+.panic:
     ud2
 
 ap_relocated_func_end:
```

td-shim/src/bin/td-shim/asm/ap_loop.asm
```diff
@@ -0,0 +1,90 @@
+# Copyright (c) 2022 Intel Corporation
```
jodh-intel, line 1: This file has Windows line endings.

jodh-intel (commented):

td-shim/src/bin/td-shim/asm/ap_loop.asm
gaojiaqi7: Right, they should be local

td-shim/src/bin/td-shim/asm/ap_loop.asm
gaojiaqi7: Thanks for reminding😊

td-layout/src/memslice.rs
```diff
@@ -95,10 +97,15 @@ pub unsafe fn get_dynamic_mem_slice_mut<'a>(t: SliceType, base_address: usize) -
             base_address as *const u8 as *mut u8,
             TD_PAYLOAD_EVENT_LOG_SIZE as usize,
         ),
+        SliceType::RelocatedMailbox => core::slice::from_raw_parts_mut(
```
jiangliu, line 100: Please also add test cases to cover this new area.

td-layout/src/runtime.rs
```diff
@@ -38,6 +40,7 @@
 
 pub const TD_PAYLOAD_EVENT_LOG_SIZE: u32 = 0x100000;
 pub const TD_PAYLOAD_ACPI_SIZE: u32 = 0x100000;
+pub const TD_PAYLOAD_MAILBOX_SIZE: u32 = 0x100000;
```
jiangliu, line 43: Why does it need 1MB? What's the minimal size for this area?

gaojiaqi7: Setting this area to 1MB is to reserve space for the AP function since it can't be allocated dynamically.
But at present, it seems that it only consumes about 100 bytes. So I think the minimal size is 4KB (mailbox page) plus size of AP function (about 100 bytes) and the proper one can be 8KB (0x2000).
What do you think of this?

td-layout/src/memslice.rs
gaojiaqi7: Sure, thanks for reminding

td-layout/src/runtime.rs
jiangliu: I think 2 pages should be ok, one for data and one for code.

jiangliu (approved):

PR #127, opened 2022-02-28
Add bounds and sanity checks in FV lib

gaojiaqi7:
- Solve the issues found by fuzz
- Add sanity check for firmware volume header

### DIFF
--- a/td-shim-tools/src/enroller.rs
+++ b/td-shim-tools/src/enroller.rs
@@ -17,7 +17,7 @@ use td_shim::secure_boot::{
     CfvPubKeyFileHeader, CFV_FILE_HEADER_PUBKEY_GUID, PUBKEY_FILE_STRUCT_VERSION_V1,
     PUBKEY_HASH_ALGORITHM_SHA384,
 };
-use uefi_pi::pi::fv::{FIRMWARE_FILE_SYSTEM3_GUID, FV_FILETYPE_RAW};
+use uefi_pi::pi::fv::{FIRMWARE_FILE_SYSTEM3_GUID, FVH_REVISION, FVH_SIGNATURE, FV_FILETYPE_RAW};
 
 use crate::public_key::{
     RsaPublicKeyInfo, SubjectPublicKeyInfo, ID_EC_PUBKEY_OID, RSA_PUBKEY_OID, SECP384R1_OID,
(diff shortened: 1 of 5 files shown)


td-shim-tools/src/enroller.rs
```diff
@@ -32,8 +32,12 @@ pub fn build_cfv_header() -> FvHeader {
         .fv_header
         .file_system_guid
         .copy_from_slice(FIRMWARE_FILE_SYSTEM3_GUID.as_bytes());
+    cfv_header.fv_header.signature = FVH_SIGNATURE;
+    cfv_header.fv_header.header_length = size_of::<FvHeader>() as u16;
     cfv_header.fv_header.fv_length = TD_SHIM_CONFIG_SIZE as u64;
-    cfv_header.fv_header.checksum = 0xdc0a;
+    cfv_header.fv_header.revision = FVH_REVISION;
+    cfv_header.fv_header.update_checksum();
```
jodh-intel, line 39: Calculating the checksum looks like an improvement to me but see my comments on this method below.

uefi-pi/src/pi/fv.rs
```diff
@@ -62,6 +76,16 @@ impl FirmwareVolumeHeader {
     pub fn as_bytes(&self) -> &[u8] {
         unsafe { &*slice_from_raw_parts(self as *const Self as *const u8, size_of::<Self>()) }
     }
+
+    // Calculate and update the checksum of the FirmwareVolumeHeader
+    pub fn update_checksum(&mut self) {
+        self.checksum = u16::MAX - sum16(self.as_bytes()) + 1;
+    }
```
jodh-intel, line 83: What worries me is that this method reads the struct as bytes and then modifies that struct to write the checksum back into it. That implies that if there is _already_ a checksum set, the new checksum calculation will be wrong.

A simple fix would be to set `fv_header.checksum = 0` before calling `update_checksum()` I think as that would effectively exclude the checksum from the calculation of... the checksum :smile:

```suggestion
    pub fn update_checksum(&mut self) {
        // Exclude any existing checksum from the checksum calculation
        self.checksum = 0;
        
        self.checksum = u16::MAX - sum16(self.as_bytes()) + 1;
    }
```

That would also make the method idempotent (it currently isn't fwics).

uefi-pi/src/pi/fv.rs
```diff
@@ -266,3 +290,25 @@ impl CommonSectionHeader2 {
         unsafe { &*slice_from_raw_parts(self as *const Self as *const u8, size_of::<Self>()) }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_fvh_checksum() {
```
jodh-intel, line 299: Could you add tests for:

- A `FirmwareVolumeHeader` with an existing checksum to ensure the checksum is ignored.
- Calling `update_checksum()` multiple times to ensure it has no effect on the calculated checksum.

jodh-intel (commented):

uefi-pi/src/pi/fv.rs
gaojiaqi7: Right, it was missed, thanks for your suggestion!😊

uefi-pi/src/fv.rs
```diff
@@ -19,15 +19,31 @@ use scroll::Pread;
 
 use crate::pi::fv::*;
 
+// Read FV header from slice and validate its integrity
+pub fn read_fv_header(fv_data: &[u8]) -> Option<FirmwareVolumeHeader> {
```
jodh-intel, line 23: Could you add some unit tests for this function?

gaojiaqi7: Sure

jyao1 (approved):

PR #130, opened 2022-03-03
Refactor enroll tool

gaojiaqi7:
- Add a self implemented type Guid for more flexible use and replace the one provided by external library.
- Extend the td-shim-enroll to support enrollment of raw files into CFV.

PR #132, opened 2022-03-07
Refact test-td-payload in tests

liuw1:
Implement a simple test framework to run test cases with TDX dependency.

This PR depends on [PR 130](https://github.com/confidential-containers/td-shim/pull/130). It will use latest enroll tool to enroll the test json file in CFV. 

### DIFF
--- a/doc/test_with_td_payload.md
+++ b/doc/test_with_td_payload.md
@@ -1,0 +1,90 @@
+# Test with TD Payload
+TD-Shim has some test cases rely on TDX specific environment. Therefore, this simple test framework is created to support the test.
+
+## Test framework code structure
+All codes of test framework are located in [tests/test-td-payload](../tests/test-td-payload).
+
+```
+-- tests
+---- test-td-payload
+------ main.rs
+------ lib.rs
+------ test.json
+------ testtdinfo.rs
+------ ...
+```
+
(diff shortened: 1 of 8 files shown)


doc/test_with_td_payload.md
```diff
@@ -0,0 +1,89 @@
+# Test with TD Payload
+TD-Shim has some test cases rely on TDX specific environment. Therefore, this simple test framework is created to support the test.
+
+## Test framework code structure
+All codes of test framework are located in [tests/td-test-payload](../tests/test-td-payload).
```
jiangliu, line 5: tests/test-td-payload?

tests/test-td-payload/Cargo.toml
```diff
@@ -1,26 +1,30 @@
 [package]
 name = "test-td-payload"
 version = "0.1.0"
-description = "Run TDX hardware specific unit test cases inside a VM"
```
jiangliu, line 4: It would be better to keep these lines.

liuw1: Agree

doc/test_with_td_payload.md
liuw1: Thank you, I will correct it.

jiangliu:
This test framework should also work on normal x86 hardware when the tdx feature is disabled, so could we test it in the  github integration workflow with tdx disabled?

liuw1:
> This test framework should also work on normal x86 hardware when the tdx feature is disabled, so could we test it in the github integration workflow with tdx disabled?

This test framework created only for testing TDX related features, it can only works when the TDX feature is enabled.

Issue #134, opened 2022-03-07
CI: Fails to build PE format payload


gaojiaqi7:
running "clang" "-O0" "-ffunction-sections" "-fdata-sections" "-g" "-fno-omit-frame-pointer" "--target=x86_64-pc-windows-msvc" "-I" "include" "-Wall" "-Wextra" "/GS" "/Gy" "/EHsc" "/GR-" "/Zc:wchar_t" "/Zc:forScope" "/Zc:inline" "/Zc:rvalueCast" "/sdl" "/Wall" "/wd4127" "/wd4464" "/wd4514" "/wd4710" "/wd4711" "/wd4820" "/wd5045" "/Od" "/RTCsu" "/WX" "-c" "/FoD:\\a\\td-shim\\td-shim\\target\\debug\\build\\ring-917f19b5a6b6e7a0\\out\\aes_nohw.obj" "crypto/fipsmodule/aes/aes_nohw.c"
  clang: error: no such file or directory: '/GS'
  clang: error: no such file or directory: '/Gy'
  clang: error: no such file or directory: '/EHsc'
  clang: error: no such file or directory: '/GR-'
  clang: error: no such file or directory: '/Zc:wchar_t'
  clang: error: no such file or directory: '/Zc:forScope'
  clang: error: no such file or directory: '/Zc:inline'
  clang: error: no such file or directory: '/Zc:rvalueCast'
  clang: error: no such file or directory: '/sdl'
  clang: error: no such file or directory: '/Wall'; did you mean '-Wall'?
  clang: error: no such file or directory: '/wd4127'
  clang: error: no such file or directory: '/wd4464'
  clang: error: no such file or directory: '/wd4514'
  clang: error: no such file or directory: '/wd4710'
  clang: error: no such file or directory: '/wd4711'
  clang: error: no such file or directory: '/wd4820'
  clang: error: no such file or directory: '/wd5045'
  clang: error: no such file or directory: '/Od'
  clang: error: no such file or directory: '/RTCsu'
  clang: error: no such file or directory: '/WX'
  clang: error: no such file or directory: '/FoD:\a\td-shim\td-shim\target\debug\build\ring-917f19b5a6b6e7a0\out\aes_nohw.obj'
  thread 'main' panicked at 'execution failed', C:\Users\runneradmin\.cargo\git\checkouts\ring-1836a5bf48f90a2a\3e04208\build.rs:663:9
[...]

Issue #109, opened 2022-02-21
Clarify why shim won't build with GCC


jodh-intel:
https://github.com/confidential-containers/td-shim/blob/staging/README.md shows that the shim must be built with `clang`. However, we should:

- Explain why the default compiler oin most Linux distros (GCC) won't work.
- Understand and ideally fix the GCC issue.

xiaoyuxlu:
> Explain why the default compiler oin most Linux distros (GCC) won't work.

My understanding is that
Rustc target of `TD-shim` is `x86_64-unknown-uefi` (PE format). Linux GCC generates ELF objs, but `rustc target x86_64-unknown-uefi` generates PE objs. So link failed. Maybe x86_64-w64-mingw32-gcc would work. 

> Understand and ideally fix the GCC issue.

I think it's very difficult at the moment. Unless:
the distribution (GCC) supports compiling the PE format. 
Or
We choose another rust-toolchain, like: x86_64-unknown-none(tier3)(elf format).  But in this way we can't solve the problem that the `cl.exe` compiler can't build the elf obj on windows. 😭 

So we choose `clang`.

ariel-adam:
@jodh-intel is this issue still relevant or can be closed? 
If it's still relevant to what release do you think we should map it to (mid-November, end-December, mid-February etc...)?

jodh-intel:
Closing since @xiaoyuxlu has explained this limitation.

PR #138, opened 2022-03-10
Modify environment variables so that CI can pass

haowqs:
Fix: #134 

Add and modify CC and AR environment variables in .github/workflows/main.yml .

Modify the required_compiler variable value in build.rs to make it pass the check .

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,9 +10,9 @@ on:
 name: RUN CODE
 
 env:
-  AR: llvm-ar
   AS: nasm
-  CC: clang
+  AR_x86_64_unknown_uefi: llvm-ar
+  CC_x86_64_unknown_uefi: clang
   RUST_TOOLCHAIN: nightly-2021-08-20
   TOOLCHAIN_PROFILE: minimal
 
--- a/td-shim/build.rs
+++ b/td-shim/build.rs
@@ -2,15 +2,13 @@
 //
 // SPDX-License-Identifier: BSD-2-Clause-Patent
 
+use anyhow::Result;
 use std::{
     env, format,
     path::{Path, PathBuf},
     process::{exit, Command},
 };
-
-use anyhow::{anyhow, Result};
 use td_layout::build_time;
-use which::which;
 
 fn nasm(file: &Path, arch: &str, out_file: &Path, args: &[&str]) -> Command {
     let oformat = match arch {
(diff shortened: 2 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -10,9 +10,10 @@ on:
 name: RUN CODE
 
 env:
-  AR: llvm-ar
   AS: nasm
-  CC: clang
+  CC_x86_64_unknown_uefi: clang
+  AR_x86_64_unknown_uefi: llvm-ar
+  ACTIONS_ALLOW_UNSECURE_COMMANDS: true
```
jiangliu, line 16: What's this flag used for?

td-shim/build.rs
```diff
@@ -43,15 +43,15 @@ fn run_command(mut cmd: Command) {
...
     const AS_ENV_VAR: &str = "AS";
-    const AR_ENV_VAR: &str = "AR";
+    const AR_ENV_VAR: &str = "AR_x86_64_unknown_uefi";
 
     // Defaults to 'cc' but also honours the CC_ENV_VAR environment variable.
     let cfg = cc::Build::new().try_get_compiler()?;
 
     // GCC cannot be used as it fails to link.
-    let required_compiler = "clang";
+    let required_compiler = "cl";
```
xiaoyuxlu, line 54: `clang` is  required to build ring when `--target` is `x86_64-unknown-uefi`).  I don't think you should change this.

.github/workflows/main.yml
```diff
@@ -28,6 +29,13 @@ jobs:
           - ubuntu-20.04
           - windows-2019
     steps:
+      # When the system is linux, add environment variables so that the compiler command is clang
+      - name: Linux add environment variables
+        if: runner.os == 'Linux'
+        run: |
+          echo '::set-env name=CC::clang'
+          echo '::set-env name=AR::llvm-ar'
+
```
xiaoyuxlu, line 38: Unnecessary. We already set `CC_x86_64_unknown_uefi: clang`

.github/workflows/main.yml
xiaoyuxlu, line 16: Unnecessary. I think we should avoid to use `ACTIONS_ALLOW_UNSECURE_COMMANDS`

xiaoyuxlu (changes requested):

.github/workflows/main.yml
xiaoyuxlu: `CC_x86_64_unknown_uefi`  supply with certain CC  for x86_64_unkown_uefi target. 
It only impact `x86_64-unknwn-uefi` target build.

| environment | build target | td-shim-tool/ring | td-shim/ring | 
| - | - | - | - |
|CC=clang | host target| clang | NA |
|CC=clang | x86_64-unknown-uefi |  NA  | clang    |
|CC_x86_64_unknown_uefi=clang | host target| host compiler(cl.exe/cc)|  NA |
|CC_x86_64_unknown_uefi=clang | x86_64-unknown-uefi |  NA  | clang    |

FYI:

`cc-rs`: the prioritized order for CC environment variables:

```
<var>_<target> - for example, CC_x86_64-unknown-linux-gnu
<var>_<target_with_underscores> - for example, CC_x86_64_unknown_linux_gnu
<build-kind>_<var> - for example, HOST_CC or TARGET_CFLAGS
<var> - a plain CC, AR as above.
```

jiangliu: It would be great to document this, it gives info about how the system works.

xiaoyuxlu: `clang` is only required when building `ring` with the `x86_64-unknown-uefi` target.  #109

xiaoyuxlu (approved):

jiangliu (approved):

PR #141, opened 2022-03-14
[Move] td-loader

haowqs:
Parser and loader for 64 bit Portable Executable (PE+) or 32/64 bit Executable and Linkable Format(ELF) binary objects.


jyao1 (approved):

td-loader/fuzz/Cargo.toml
```diff
@@ -0,0 +1,32 @@
+[package]
```
jiangliu, line 1: Should we move fuzz related code into a dedicated directory such as `td-shim/fuzz`?
And it would be better to open a PR for the fuzz test code.

jyao1: We can have 2 model:

1) fuzz under feature.
```
<Root>
---- <Feature1>
-------- <Fuzz>
---- <Feature2>
-------- <Fuzz>
```

2) feature under fuzz
```
<Root>
---- <Feature1>
---- <Feature2>
---- <Fuzz>
-------- <Feature1>
-------- <Feature2>
```

We are using option 1 for cargo-fuzz and option 2 for afl.rs. (Unfortunately)
Long term, I think we need align to option 1), to support cargo-fuzz. Any concern on that?

jiangliu: Seems cargo-fuzz generates a dedicated workspace, is it a must to host the fuzz test crate inside the target crate?
Personally I prefer option 2, because we may have dedicated directories for tests/benchmarks/fuzz tests etc.

jyao1: Good point. My thought is:
* The system level tests/benchmarks/fuzz will be under root dir.
* The unit level tests or fuzz should be in feature dir.

jiangliu: Great to classify fzz tests into two classes: unit level or system level.
That's clear now.

PR #143, opened 2022-03-15
[Move] td-uefi-pi is migrating main branch

haowqs:
Migrate the staging branch uefi-pi to the main branch td-uefi-pi

### DIFF
--- a/td-uefi-pi/src/fv.rs
+++ b/td-uefi-pi/src/fv.rs
@@ -19,15 +19,31 @@ use scroll::Pread;
 
 use crate::pi::fv::*;
 
+// Read FV header from slice and validate its integrity
+pub fn read_fv_header(fv_data: &[u8]) -> Option<FirmwareVolumeHeader> {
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -10,7 +10,7 @@ members = [
 	"td-layout",
 	"td-logger",
 	"td-paging",
-	"td-uefi-pi",
```
jiangliu, line 13: It would be better to keep it as `td-uefi-pi`, so we have a common prefix for all crates

jiangliu:
The crate is specially designed for td, so it would be better to keep it as `td-uefi-pi`.
And the r-uefi-pi should already have been merged into td-uefi-pi, so just need to merge several PRs from staging into main.

Cargo.toml
haowqs: Done

haowqs:
I changed the commit info.

jyao1 (approved):

jiangliu (approved):

PR #146, opened 2022-03-15
[Move] td-logger: Minor cleanups for td-logger crate

haowqs:

### DIFF
--- a/td-logger/src/logger.rs
+++ b/td-logger/src/logger.rs
@@ -2,28 +2,39 @@
 //
 // SPDX-License-Identifier: BSD-2-Clause-Patent
 
-//! A standalone logger to directly log messages, independent of the log crate.
-
-use core::fmt::{self, Write};
+use core::fmt;
 use lazy_static::lazy_static;
 use spin::Mutex;
 
 use crate::{dbg_port_write, dbg_write_string};
 
-pub const LOG_LEVEL_TRACE: usize = 5;
-pub const LOG_LEVEL_DEBUG: usize = 4;
-pub const LOG_LEVEL_INFO: usize = 3;
-pub const LOG_LEVEL_WARN: usize = 2;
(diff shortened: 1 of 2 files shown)


jyao1:
Is that really clean up?

I feel the main is correct.

td-logger/src/logger.rs
```diff
@@ -2,28 +2,39 @@
...
 pub const LOG_LEVEL_NONE: usize = 0;
 
 pub const LOG_MASK_COMMON: u64 = 0x1;
+// Core - Boot Service (BIT1 ~ BIT15)
+pub const LOG_MASK_PROTOCOL: u64 = 0x2;
+pub const LOG_MASK_MEMORY: u64 = 0x4;
+pub const LOG_MASK_EVENT: u64 = 0x8;
+pub const LOG_MASK_IMAGE: u64 = 0x10;
+// Core - Runtime Service (BIT16 ~ BIT 23)
+pub const LOG_MASK_VARIABLE: u64 = 0x10000;
```
jyao1, line 24: Why we need add those?
I think main is correct.

jyao1:
@jiangliu , I think we run into the case, that main branch fixed the issue, but staging branch does not.
Then people just gets confused on which one is correct. 

I think before we fully migrate to main, any fix in main must also be present in staging.

jiangliu:
> @jiangliu , I think we run into the case, that main branch fixed the issue, but staging branch does not. Then people just gets confused on which one is correct.
> 
> I think before we fully migrate to main, any fix in main must also be present in staging. To eliminate the confusing, should be drop this patch, then create another one to merge the fix from trunk to staging?

I think we should do that, eventually we should merge the staging into the main branch to keep the development history.

jyao1 (changes requested):
Would you please check which one the latest? Main or staging?
Now I feel main is right. If so, we need back port from main to staging.

PR #148, opened 2022-03-15
[Move] tdx-tdcall: minor cleanups

haowqs:


jyao1 (approved):

jyao1 (changes requested):
Would you please check which one the latest? Main or staging?
Now I feel main is right. If so, we need back port from main to staging.

jiangliu:
> Would you please check which one the latest? Main or staging? Now I feel main is right. If so, we need back port from main to staging.

yes, seems we should merge the reverse patch from main into staging.

PR #151, opened 2022-03-16
[Move] td-shim-tools: Migrate staging branch td-shim-tools to main branch

haowqs:


jiangliu:
It would be better to split it into function units:
1) ld to generate the final binary
2) secure-boot related code and doc

jyao1 (approved):

jyao1:
If no more comment, I will merge.

PR #161, opened 2022-03-21
[Move] doc: Migrate staging markdown files to main

haowqs:


jiangliu:
It would be better to submit documentation with associated code. For example, the “design.md" may be committed with the td-shim code.

PR #164, opened 2022-03-21
test-td-payload: Add ACPI test case

mxu9:
This is the initial version of ACPI test case. In this test case ACPI
tables will be checked if they exist in payload hob. In the next version
more check will be executed to the ACPI tables, such as the validness of
paticular fields.

Signed-off-by: Min Xu <min.m.xu@intel.com>

### DIFF
--- a/tests/test-td-payload/src/test.json
+++ b/tests/test-td-payload/src/test.json
@@ -63,5 +63,27 @@
         "name": "testve",     
         "result": "None",
         "run": true
+    },
+    "tcs008": {
+        "name": "testacpi",
+        "expected": {
+            "num": 2,
+            "tables":[
+                {
+                    "name": "APIC",
+                    "signature": [65, 80, 73, 67],
+                    "valid": 1,
+                    "exist": 1
+                },
+                {
+                    "name": "TDEL",
+                    "signature": [84, 68, 69, 76],
+                    "valid": 1,
+                    "exist": 1
+                }
+            ]
+        },
+        "result": "None",
+        "run": true
     }
 }
--- a/tests/test-td-payload/src/testacpi.rs
+++ b/tests/test-td-payload/src/testacpi.rs
@@ -1,0 +1,179 @@
+// Copyright (c) 2021 Intel Corporation
+//
+// SPDX-License-Identifier: BSD-2-Clause-Patent
+
+#![no_std]
+extern crate alloc;
+
+use crate::lib::{TestCase, TestResult};
+use alloc::string::{String, ToString};
+use alloc::vec::Vec;
+use core::ffi::c_void;
+use core::mem::size_of;
+use serde::{Deserialize, Serialize};
+use td_layout::memslice;
+use td_shim::acpi::GenericSdtHeader;
+use td_shim::TD_ACPI_TABLE_HOB_GUID;
+use td_uefi_pi::hob;
+use zerocopy::FromBytes;
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct TdAcpiData {
+    pub name: String,
+    pub signature: [u8; 4],
+    pub valid: u8,
+    pub exist: u8,
+}
+
+impl Default for TdAcpiData {
+    fn default() -> TdAcpiData {
+        TdAcpiData {
+            name: String::default(),
+            signature: [0, 0, 0, 0],
+            valid: 0,
+            exist: 0,
+        }
+    }
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct TdAcpiList {
+    pub num: u32,
+    pub tables: [TdAcpiData; 2],
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct TdAcpi {
+    pub name: String,
+    pub expected: TdAcpiList,
+    pub result: TestResult,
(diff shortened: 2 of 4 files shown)


tests/test-td-payload/src/test.json
```diff
@@ -34,5 +37,29 @@
         },
         "result": "None",
         "run": false 
-    }        
-}
\ No newline at end of file
+    },
+    "tcs004": {
+        "name": "tdacpi001",
+        "hob": 0,
```
jyao1, line 43: why we need add hob with all zero data?

tests/test-td-payload/src/testtdinfo.rs
```diff
@@ -27,6 +27,7 @@ pub struct TdInfoRetData {
 #[derive(Debug, Serialize, Deserialize)]
 pub struct Tdinfo {
     pub name: String,
+    pub hob: usize,
```
jyao1, line 30: hob is runtime data. Why mix it with static data?

tests/test-td-payload/src/testtdinfo.rs
```diff
@@ -27,6 +27,7 @@ pub struct TdInfoRetData {
 #[derive(Debug, Serialize, Deserialize)]
 pub struct Tdinfo {
```
jyao1, line 28: the name is confusing. Should it be TdTestInfo?

tests/test-td-payload/src/test.json
mxu9: The function build_testcases (which is at tests/test-td-payloader/src/main.rs) parse the test.json and generate the TestCases object. TestCases contains the TdAcpi object which is parsed from test.json. TdAcpi need the information of hob so that it can parse hob to get the ACPI tables. The actual value of hob is passed to TdAcpi before it is started to run.
In test.json hob field is a place holder so that build_testcases can generate the TdAcpi object.

tests/test-td-payload/src/testtdinfo.rs
mxu9: hob is not needed in Tdinfo. It will be removed.

tests/test-td-payload/src/testtdinfo.rs
mxu9: The name is not related to this PR. I will ping Wei about it.

tests/test-td-payload/src/test.json
mxu9: The field of "hob" means the base address of hob list. The name is confusing. It has been updated to hob_address in the new version.

liuw1 (approved):

tests/test-td-payload/src/test.json
jyao1: But I don't understand why we need input hob_address in jason file.

Why does a test case writer need aware there is hob_address field and need initialize to zero?

mxu9: Not all the test case need a hob_address field. For example Hob address is not needed in Testcase of TdInfo. 
But there is some testcase need be aware of hob_address. TdAcpi is the example. Because ACPI tables are passed in hob_list and the Testcase of TdAcpi parses hob_list to check if the tables exist and if the tables are correct.
So in the new version of test.json only tcs004 has the field of hob_address.

test.json describes the test cases. But hob_address is unknown to the test.json. So the initial value of hob_address is set to 0. The actual value will be set when TdAcpi is added to testsuites.

jyao1: > But hob_address is unknown to the test.json.

That is exactly my point. If hob_address is unknown to a case developer, what is the value to set to 0 in test.json?
It should be totally invisible to case developer.

liuw1: It's the test framework limitation, json data should mapping with test case structure. So if min want to use hob, he need add hob_address in json file. Then json data can be parsed successfully.

liuw1 (approved):
Looks good to me! Thanks Min

jyao1:
Would you please resolve conflict?

mxu9:
> Would you please resolve conflict?

The conflicts are resolved.

tests/test-td-payload/src/testacpi.rs
```diff
@@ -0,0 +1,179 @@
...
+use alloc::string::{String, ToString};
+use alloc::vec::Vec;
+use core::ffi::c_void;
+use core::mem::size_of;
+use serde::{Deserialize, Serialize};
+use td_layout::memslice;
+use td_shim::acpi::GenericSdtHeader;
+use td_shim::TD_ACPI_TABLE_HOB_GUID;
+use td_uefi_pi::hob;
+use zerocopy::{AsBytes, FromBytes};
```
jyao1, line 18: Why we need this?

mxu9: GenericSdtHeader::read_from is used to extract the ACPI header from hob_list. It is derived from zerocopy::FromBytes. 
zerocopy::AsBytes is not needed. It will be removed.

mxu9: read_from is implemented in the trait of zerocopy::FromBytes. So we must import zerocopy::FromBytes in testacpi.rs.

jyao1 (approved):

PR #181, opened 2022-03-27
fix the regression on frame allocator for page table

gaojiaqi7:
Correct the miatake use of 'base' field of BMFrameAllocator, otherwise the
allocated frames will be started from a wrong value. The page table only
occupies 1M bytes memory at present, so the problem has not been exposed
yet.

In addition, a page must be reserved for level-4 table when we init the
frame allocator.


jyao1 (approved):

jyao1:
merge regression issue to unblock the progress.

PR #190, opened 2022-03-28
Fix hardcode integrity_check value of FfsHeader

### LINKED ISSUE #170: [Bug] The 'integrity_check' of 'FfsFileHeader' should not be an hardcode value
The 'integrity_check' of 'FfsFileHeader' should not be a hardcode value. Refer to Platform Initialization Specification, Vol. 3 Chapter 3.2.3, it should be a checksum value.

Eg.
Line 64 ```integrity_check: 0xaae4,``` in td-shim/src/fv.rs.

liuw1:
1. Remove hardcode implementation and auto generate value of
integrity_check with update_checksum().
2. Validate integrity_check
3. Correct integrity_check values of files(cfv and fv_buffer).

Fix #170 


td-shim-tools/src/enroller.rs
```diff
@@ -26,6 +26,21 @@ use crate::public_key::{
...
 use crate::{write_u24, InputData, OutputFile};
 
+fn update_checksum(data: &mut [u8]) {
+    // Clear checksum bit to zero
+    data[16] = 0x0u8;
+
+    let mut sum = 0x0u8;
+    for i in 0..24 {
+        if i == 17 || i == 23 {
+            continue;
```
jyao1, line 36: please dont hardcode any value.

jyao1 (changes requested):
Please input a FFS data structure for fn update_checksum(), and remove any hardcoded number.

td-shim-tools/src/enroller.rs
liuw1: Removed all the hardcode implementation.

PR #195, opened 2022-03-31
Modify fuzz

haowqs:
- Cleanup the td-loader and td-uefi-pi fuzz
  remove  redundant code
- Modify the fuzzing.sh
  update afl run path where valid coverage report can be generated
- Update fuzzing.md
  rust toolchain update, update related content

Signed-off-by: haowei <WeiX.Hao@intel.com>

### DIFF
--- a/doc/fuzzing.md
+++ b/doc/fuzzing.md
@@ -31,13 +31,13 @@ $ sudo mount -t tmpfs -o size=1024M tmpfs in
 
 ### Build the fuzz target
 
-`$ cd td-loader/fuzz`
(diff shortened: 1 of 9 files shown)


jyao1:
Why we need move the location?

Please submit an issue at first with a RFC proposal at first.

jyao1:
please resolve the conflict.

jyao1:
please resolve the conflict.

jiangliu (approved):

haowqs:
> @haowqs, Why we need move the location?
> 
> Please submit an issue at first with a RFC proposal at first.

There's no reason to move, so don't move

PR #197, opened 2022-04-02
Add memory map test case

gaojiaqi7:


tests/test-td-payload/src/testmemmap.rs
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    /**
+     * run the test case
+     */
+    fn run(&mut self) {
+        if let Some(e820_table) = self.parse_hob(self.hob_address) {
+            self.case.result = self.verify_memory_map(e820_table);
+        } else {
+            self.case.result = TestResult::Fail;
```
liuw1, line 139: It would be better to add detailed log message when no e820_table can be found.

tests/test-td-payload/src/testmemmap.rs
```diff
@@ -0,0 +1,161 @@
...
+            {
+                return TestResult::Fail;
+            }
+            total += entry.size;
+            top = entry_end;
+        }
+
+        if let Some(memory_size) = self.case.expected.get_size() {
+            if total != memory_size {
+                TestResult::Fail
```
liuw1, line 110: Add detailed log message for line 102 and line 110 would be better.

jyao1 (approved):

liuw1 (approved):

Issue #174, opened 2022-03-24
Implement function to pass ACPI tables to TD payload


liuw1:

liuw1:
#244

Issue #94, opened 2022-02-14
regression: td_payload: New assert in td-payload/src/main.rs", line: 103


xiaoyuxlu:
This issue is found in commit hash 3a3bfe5fbed2a1c942283fd9a8fcff2be6969514

Run the latest code, it will show below exception message:
```
INFO - Starting td-payload hob - 0x3fa00000
INFO - setup_exception_handlers done
INFO - PhitHob:
INFO - version - 0x9
INFO - boot_mode - 0x0
INFO - efi_memory_top - 0x0000000040000000
INFO - efi_memory_bottom - 0x000000003d000000
INFO - efi_free_memory_top - 0x0000000040000000
INFO - efi_free_memory_bottom - 0x000000003d001000
INFO - efi_end_of_hob_list - 0x000000003fa00158
INFO - FirmwareVolume : 0x00000000ff084000 - 0x00000000ffcaefff
INFO - Cpu : mem size 36 , io size 16
INFO - MemoryAllocation 0x00000003 : 0x0000000001100000 - 0x0000000001d2afff
INFO - MemoryAllocation 0x00000004 : 0x0000000000800000 - 0x00000000008fffff
INFO - MemoryAllocation 0x00000004 : 0x000000003f000000 - 0x000000003f7fdfff
INFO - ResourceDescription 0x00000000 : 0x0000000000100000 - 0x000000003fffffff (0x00003c07)
INFO - ResourceDescription 0x00000000 : 0x0000000000000000 - 0x000000000009ffff (0x00003c07)
INFO - panic ... PanicInfo { payload: Any { .. }, message: Some(assertion failed: `(left == right)`
left: `1023410176`,
right: `1040187392`), location: Location { file: "td-payload/src/main.rs", line: 103, col: 9 } }
```

gaojiaqi7:
Fixed by: https://github.com/confidential-containers/td-shim/commit/88d1bf22024d3db5e848bb8b21d4bd10d61e3a75

PR #198, opened 2022-04-03
Handoff ACPI and E820 to executable(PE/ELF) payload through HOB

gaojiaqi7:
Fix below issues:
https://github.com/confidential-containers/td-shim/issues/174
https://github.com/confidential-containers/td-shim/issues/94

### DIFF
--- a/td-payload/src/main.rs
+++ b/td-payload/src/main.rs
@@ -17,6 +17,17 @@
 #![cfg_attr(not(test), no_main)]
 #![allow(unused)]
 
+use core::mem::size_of;
+
+use alloc::vec::Vec;
+use scroll::{Pread, Pwrite};
+use td_layout::RuntimeMemoryLayout;
+use td_shim::e820::{E820Entry, E820Type};
+use td_shim::{TD_ACPI_TABLE_HOB_GUID, TD_E820_TABLE_HOB_GUID};
+use td_uefi_pi::hob;
+use td_uefi_pi::pi;
+use zerocopy::FromBytes;
+
 #[macro_use]
 extern crate alloc;
 
@@ -31,6 +42,9 @@ extern "win64" {
     #[cfg(feature = "cet-ss")]
     fn cet_ss_test(count: usize);
 }
+
+const E820_TABLE_SIZE: usize = 128;
+const PAYLOAD_HEAP_SIZE: usize = 0x400_0000;
 
 #[cfg(not(test))]
 mod payload_impl {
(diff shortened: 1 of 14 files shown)


td-payload/src/main.rs
```diff
@@ -34,7 +44,95 @@ extern "win64" {
     fn cet_ss_test(count: usize);
 }
 
-#[cfg(not(test))]
+pub struct PayloadHobInfo<'a> {
+    hob_list: &'a [u8],
+    page_table: pi::hob::MemoryAllocationHeader,
+    payload: pi::hob::MemoryAllocationHeader,
+    runtime: pi::hob::MemoryAllocationHeader,
```
jyao1, line 51: Do we assume only one runtime memory block is available?

jyao1:
please resolve conflict.

jyao1 (approved):

td-shim/src/lib.rs
```diff
@@ -6,21 +6,55 @@
 
 use scroll::{Pread, Pwrite};
 
-use td_uefi_pi::pi;
+use td_uefi_pi::{pi, pi::guid};
```
jiangliu, line 9: You may also use use td_uefi_pi::pi::{self, guid};

td-shim/src/lib.rs
```diff
@@ -6,21 +6,55 @@
...
 pub mod secure_boot;
 
-pub const TD_ACPI_TABLE_HOB_GUID: [u8; 16] = [
-    0x70, 0x58, 0x0c, 0x6a, 0xed, 0xd4, 0xf4, 0x44, 0xa1, 0x35, 0xdd, 0x23, 0x8b, 0x6f, 0xc, 0x8d,
-];
-
-pub const TD_KERNEL_INFO_HOB_GUID: [u8; 16] = [
-    0x12, 0xa4, 0x6f, 0xb9, 0x1f, 0x46, 0xe3, 0x4b, 0x8c, 0xd, 0xad, 0x80, 0x5a, 0x49, 0x7a, 0xc0,
-];
+pub const TD_ACPI_TABLE_HOB_GUID: guid::Guid = guid::Guid::from_fields(
```
jiangliu, line 17: Could you please help to add documentation for related sections in the td-shim spec?

td-payload/src/main.rs
gaojiaqi7: The 'runtime' indicates the whole region located at the highest memory below 4G which contains heap, stack, evenlog and ACPI table etc., in runtime layout.

Do you think it is good to name this block as highmem or reserved?

jyao1: why we need one specific block to indicate highmem?

td-payload cannot make any assumption on td-shim, if it is not defined in td-shim spec.

jyao1 (changes requested):
need remove 'runtime' from `PayloadHobInfo`

td-shim/src/lib.rs
gaojiaqi7: Thanks for reminding me

td-shim/src/lib.rs
gaojiaqi7: I added some references to the spec

gaojiaqi7:
> need remove 'runtime' from `PayloadHobInfo`

hi @jyao1, 'runtime' has been removed and payload does not depend on interfaces that is not defined in td-shim spec.

gaojiaqi7:
Move to main, close this one

Issue #81, opened 2022-02-10
Implement shim loader in td-shim-tools


jiangliu:
We have a linker `td-shim-ld` in the td-shim-tools crate, there should be a loader too.
The loader should take the responsibilities:
1) parse td-shim binary file into sections
2) validate section information
3) enforce present sections follow the td-shim spec

jyao1:
1) Create a `td-shim-loader` library, that could be reused by the VMM
2) Create a `td-shim-checker` tool (consuming `td-shim-loader`), to validate the format of the final binary.

jiangliu:
> 1. Create a `td-shim-loader` library, that could be reused by the VMM
> 2. Create a `td-shim-checker` tool (consuming `td-shim-loader`), to validate the format of the final binary.

Yes, exactly.
It may be included into the td-shim-tools crate.

mxu9:
As discussed with Jiewen, I will take this task and plan to start this work after I finish test-acpi-payload. ETA is ww15.

mxu9:
> parse td-shim binary file into sections
td-shim is of PE format. Here section means the PE section?

jyao1:
It is meta section, not PE section.

jyao1:
done

PR #200, opened 2022-04-06
Td shim checker

mxu9:
REF: https://github.com/confidential-containers/td-shim/issues/81


td-shim-tools/src/loader.rs
```diff
@@ -0,0 +1,149 @@
...
+        if read_from_file(&mut file, metadata_offset as u64, &mut buffer).is_err() {
+            error!("Failed to read metadata descriptor from file");
+            return None;
+        }
+        let metadata_descriptor = TdxMetadataDescriptor::from_bytes(&buffer);
+        if metadata_descriptor.is_none() {
+            error!("Invalid TdxMetadataDescriptor");
+            return None;
+        }
+        let metadata_descriptor = metadata_descriptor.unwrap();
```
gaojiaqi7, line 102: You can use or_else() here like:
```
let metadata_descriptor = TdxMetadataDescriptor::from_bytes(&buffer).or_else(|| {
        error!("Invalid TdxMetadataDescriptor");
        None
    })?;
```
Similar methods can be used to handling errors

jyao1:
please fix CI failure

jyao1:
please fix CI failure

jyao1:
What issue this PR is trying to resolve? Would you please add "FIX" in the description?

td-layout/src/metadata.rs
```diff
@@ -80,32 +121,125 @@ pub struct TdxMetadataSection {
...
+            TDX_METADATA_SECTION_TYPE_TEMP_MEM => {
+                valid = self.data_offset == 0
+                    && self.raw_data_size == 0
+                    && self.memory_address != 0
+                    && self.memory_data_size != 0
+                    && self.attributes == 0;
+            }
+
+            TDX_METADATA_SECTION_TYPE_PAYLOAD => {
+                valid = self.data_offset == 0
```
jyao1, line 185: I think it is legal to have payload inside of binary. In this case data_offset != 0, raw_data_size != 0

td-layout/src/metadata.rs
```diff
@@ -80,32 +121,125 @@ pub struct TdxMetadataSection {
...
+                    && self.memory_data_size != 0
+                    && self.attributes == 0;
+            }
+
+            TDX_METADATA_SECTION_TYPE_PAYLOAD_PARAM => {
+                valid = self.data_offset == 0
+                    && self.raw_data_size == 0
+                    && self.memory_address != 0
+                    && self.memory_data_size != 0
+                    && self.attributes == 0;
```
jyao1, line 197: May I know what rule you are using to perform the check ?

I think we need follow Td-Shim spec,  "Rules for the TDVF_SECTION".  Please make sure the check is aligned with it.

td-shim-tools/src/loader.rs
mxu9: Thanks for the reminder. It will be updated.

mxu9:
> What issue this PR is trying to resolve? Would you please add "FIX" in the description?
Sure. Below description will be added in "FIX" section.

This PR corresponds to issue #81 which tries to resolve:
1. Parse td-shim binary to find out the TdxMetadata
2. Validate TdxMeatadata information and ensure it follows the td-shim spec

So we create one library and one tool.
1. Create a td-shim-loader library, that could be reused by the VMM. This library exports below function:
    TdShimLoader::parse(filename: &String) -> Option<(TdxMetadataDescriptor, Vec<TdxMetadataSection>)>

2. Create a td-shim-checker tool (consuming td-shim-loader), to validate the format of the final binary. 
   cargo run -p td-shim-tools --bin td-shim-checker --no-default-features --features="loader" -- target/x86_64-unknown-uefi/release/final-pe.bi

In the meantime there are some other changes in other crates. They're in td-uefi-pi and td-layout. These changes are all related to td-shim-loader and td-shim-checker. So they're in the same PR.

jyao1:
@jiangliu , would you please help to review this?

mxu9:
@jiangliu Would you please help to review this?

PR #210, opened 2022-04-08
[Move] Update rust toolchain version to nightly-2022-04-07

haowqs:
Changes:
    - global_asm and asm macro are already stable features and can be used
      by import core::arch::global_asm and core::arch::asm
    - Update crate x86, x86_64, bootloader and uart to latest version.
    - Add #![feature(asm_sym)] to resolve compilation error.
    - Update rust toolchain version used by CI.


jyao1 (approved):

.github/workflows/format.yml
```diff
@@ -29,7 +29,7 @@ jobs:
         uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: 1.58.1
+          toolchain: nightly-2022-04-07
```
jiangliu, line 32: Is this change needed? I hope we could achieve:
1) build components which will run inside tdx with nightly
2) build components which will run on host side with stable

haowqs: ok, changed it back

PR #213, opened 2022-04-12
format.yml: The rust version uses stable 1.58.1

haowqs:
- build components which will run inside tdx with nightly
- build components which will run on host side with stable

Signed-off-by: haowei <WeiX.Hao@intel.com>


jyao1:
please resolve CI failure.

PR #221, opened 2022-04-13
Add integration script and workflow for tdx server

arronwy:
The integtation script is implemented by Wei.

Co-authored-by: Wei Liu <wei3.liu@intel.com>
Signed-off-by: Arron Wang <arron.wang@intel.com>

### DIFF
--- a/.github/workflows/integration-tdx.yml
+++ b/.github/workflows/integration-tdx.yml
@@ -1,0 +1,54 @@
+on:
+  push:
+    paths-ignore:
+      - "**.md"
+  pull_request:
+    paths-ignore:
+      - "**.md"
+  workflow_dispatch:
+
+name: Ingetration Test on TDX Server
+
+env:
+  AS: nasm
+  AR_x86_64_unknown_uefi: llvm-ar
+  CC_x86_64_unknown_uefi: clang
+  RUST_TOOLCHAIN: nightly-2022-04-07
+  TOOLCHAIN_PROFILE: minimal
+
+jobs:
+  system_compile:
+    name: Run TDX Integration Test
+    runs-on: [self-hosted, tdx]
+    timeout-minutes: 30
+
+    steps:
+      # Install first since it's needed to build NASM
+      - name: Install LLVM and Clang
+        uses: KyleMayes/install-llvm-action@v1
+        with:
+          version: "10.0"
+          directory: ${{ runner.temp }}/llvm
+
+      - name: Checkout sources
+        uses: actions/checkout@v2
+
+      - name: Install toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: ${{ env.TOOLCHAIN_PROFILE }}
+          toolchain: ${{ env.RUST_TOOLCHAIN }}
+          override: true
+          components: rust-src, llvm-tools-preview
+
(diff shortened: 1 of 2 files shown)


jyao1:
> The integtation script is implemented by Wei.
> 
> Co-authored-by: Wei Liu [wei3.liu@intel.com](mailto:wei3.liu@intel.com) 

If it is implemented by Wei, I think we need: Signed-off-by: Wei Liu

arronwy:
> > The integtation script is implemented by Wei.
> > Co-authored-by: Wei Liu [wei3.liu@intel.com](mailto:wei3.liu@intel.com)
> 
> If it is implemented by Wei, I think we need: Signed-off-by: Wei Liu Also, you may use multiple signed-off, if you modified it.

Yes, fixed in commit messages

.github/workflows/integration-tdx.yml
```diff
@@ -0,0 +1,62 @@
...
+      - name: Build TdShim
+        uses: actions-rs/cargo@v1
+        with:
+          command: xbuild
+          args: -p td-shim --target x86_64-unknown-uefi --release --features=main,tdx
+
+      - name: Build PE format payload
+        run: |
+          cargo xbuild -p td-payload --target x86_64-unknown-uefi --release --features=main,tdx
+          cargo run -p td-shim-tools --features=td-shim/default --bin td-shim-ld -- target/x86_64-unknown-uefi/release/ResetVector.bin target/x86_64-unknown-uefi/release/td-shim.efi target/x86_64-unknown-uefi/release/td-payload.efi -o target/x86_64-unknown-uefi/release/final-pe.bin
```
fidencio, line 59: @arronwy, if https://github.com/confidential-containers/td-shim/pull/223 gets merged before this one, I´d  consider changing these 2 lines to, instead, just doing
```
          bash sh_script/build_final.sh pe_boot_kernel
```

That would ensure that any changes that need to be done in the future, are only done as part of that script.

.github/workflows/integration-tdx.yml
```diff
@@ -0,0 +1,62 @@
...
+
+      - name: Run Tests
+        run: bash sh_script/integration_tdx.sh -f target/x86_64-unknown-uefi/release/final-pe.bin
```
fidencio, line 62: Then it'd require changing the binary name here to `final-pe-boot-kernel.bin` instead.

fidencio (commented):
@arronwy, I've made some suggestions as your work can benefit of a series that I submitted Yesterday.

Please, if possible, consider those.

.github/workflows/integration-tdx.yml
arronwy: good suggestion, we can use this script to avoid duplicate and reduce the maintenance effort.

arronwy: @fidencio thanks, it works great!.

jiangliu:
As we are switching from staging to main, should we target this PR against the main branch?

jyao1:
> As we are switching from staging to main, should we target this PR against the main branch?

Yes. The patch will be applied to main as well.

jyao1 (approved):

jyao1:
Drop this because staging is no longer needed.
Patch for main is at https://github.com/confidential-containers/td-shim/pull/237.

PR #223, opened 2022-04-13
actions: Add an action for automatically doing releases and publishing the needed binaries

fidencio:
This series is a pre-req for being able to release the  TD-shim binaries as part of a tagged release.

This groundwork leverages the `build_final.sh` script, which currently builds the `final-pe.bin` and the `final-elf.bin` binaries, to also build a `final-pe-boot-kernel.bin` binary.

The new binary is basically the good and well known `final-pe.bin`, but built with the `boot-kernel` feature enabled, which is essential for using td-shim with Cloud Hypervisor.

Once this series is merged, we'll use this script as part of a GitHub action that will run once a release is tagged, generating then those binaries and attaching those as artefacts of the release.  The action will come on a different PR as it'll hit the `main` branch directly, while this series here targets the staging branch.

@jyao1, please, take a look at those simple changes and let me know if you have any concern about them.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,91 @@
+name: Release TD-shim (staging)
+on: create
+
+env:
+  RUST_TOOLCHAIN: nightly-2022-04-07
+  TOOLCHAIN_PROFILE: minimal
+
+jobs:
+  release:
+    if: github.event_name == 'create' && github.event.ref_type== 'tag'
+    name: Release
+    runs-on: ubuntu-20.04
+    timeout-minutes: 30
+
+    steps:
+      - name: install NASM
+        uses: ilammy/setup-nasm@v1
+
+      - name: Install LLVM and Clang
+        uses: KyleMayes/install-llvm-action@v1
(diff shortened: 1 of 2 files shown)


fidencio:
If you want to sneak-peak how the release action looks like, you can find it [here](https://github.com/fidencio/td-shim/commit/c19a18b6a9956579e167f7510d6bdb366b67106f), but do so just for the satisfaction of your curiosity, as that still requires some testing and, once done, a PR against the `main` branch will be opened.

jyao1 (approved):

fidencio:
@jyao1, I've added one more commit here as I realised that the GitHub action must also be in the `staging` branch.
So, I've dismissed your approval to avoid misleading @jiangliu. :-)

Please, @jyao1 and @jiangliu, take a look at the series.  @jyao1, the latest commit is the only difference from what you approved.

One example of the "release" workflow can be seen [here](https://github.com/fidencio/td-shim/releases/tag/0.0.0.4).

jiangliu (approved):

PR #257, opened 2022-05-03
Refactor #VE handler

gaojiaqi7:
Guest TD needs to retrieve the information when a #VE is injected by TDX
Module, otherwise TDX Module will inject a #DF next time it attempting to
inject a #VE.

### DIFF
--- a/tdx-tdcall/src/tdx.rs
+++ b/tdx-tdcall/src/tdx.rs
@@ -2,6 +2,7 @@
 //
 // SPDX-License-Identifier: BSD-2-Clause-Patent
 
+use core::mem::size_of;
 use core::result::Result;
 use core::sync::atomic::{fence, Ordering};
 use lazy_static::lazy_static;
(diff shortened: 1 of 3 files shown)


tdx-tdcall/src/tdx.rs
```diff
@@ -141,30 +125,16 @@ pub fn tdvmcall_io_read_32(port: u16) -> u32 {
...
-            core::ptr::null_mut(),
-        )
-    };
-    if ret != TDVMCALL_STATUS_SUCCESS {
-        tdvmcall_halt();
-    }
-}
-
-pub fn tdvmcall_io_write_32(port: u16, byte: u32) {
+// Perform IO write via TDVMCALL Instruction.IO leaf function
```
jiangliu, line 128: I still prefer the "read_u8/u16/u32" formal which has better readability.

jiangliu (approved):

PR #264, opened 2022-05-09
td-loader: load program header table into memory

gaojiaqi7:


gaojiaqi7:
@jyao1 do you think it is reasonable to put the payload base address as a parameter when calling the payload entry point? If it is good, then should we clarify in the spec what parameters (e.g. HOB address, payload base) are passed to the entry point？

jyao1 (approved):

PR #268, opened 2022-05-15
Reproducible build: introduce a tool to help reproducible build

### LINKED ISSUE #256: [Bug] td-shim should support reproducible build
There are some differences between 2 times build with same code base. Currently td-shim does not support reproducible build.

image

longlongyang:
fix #256

1. Introduce a reproducible tool, td-reproducible-tool, to help reproducible
2. Add strip = "symbols" to cargo.toml in root workspace to remove symbols

### DIFF
--- a/td-reproducible-tool/readme.md
+++ b/td-reproducible-tool/readme.md
@@ -1,0 +1,62 @@
+# td-reproducible-tool
+
+## Background
+
+Elements break reproducible build (x86_64-unknown-uefi + target x86_64-unknown-none) <br>
+ <br>
+```
+1. Debug related sections (PE + ELF) 
+2. Code base path like \home\USERNAME\rust-td-payload\xxx\xxx.rs (PE + ELF) 
+3. CARGO_HOME like \home\USERNAME\.cargo\xxx\xxx.rs (PE + ELF) 
+4. RUSTUP_HOME like \home\USERNAME\.rustup\xxx\xxx.rs (PE + ELF) 
+5. TimeDateStamp field in COFF file header (PE only) 
+```
+
+Solution: <br>
+
+```
+1. Use strip = "symbols" option to [profile.release] 
+2. Use strip = "symbols" option to [profile.release] 
+3. Use a post-build tool to zero CARGO_HOME string 
+4. Use a post-build tool to zero RUSTUP_HOME string
+5. Use a post-build tool to zero TimeDateStamp for PE
+```
+NOTE: 3 and 4 may be resolved directly after rust-lang RFC 3127. This tool provide a temp solution to solve 3 and 4 before RFC is implemented.
+
+## tool usage
+
+```
+td-reproducible-tool [OPTIONS]
+
+TD REPRODUCIBLE TOOL
+
+Optional arguments:
+  -h,--help             Show this help message and exit
+  -w,--workspace WORKSPACE
+                        Where to find the target folder.
+  -n,--name NAME        Name for the compiled binary.
+  -t,--target TARGET    The built target to find.
+  -p,--profile PROFILE  The built profile to find.
+  -c,--cargo_home CARGO_HOME
+                        The cargo home. If not specify, system variable CARGO_HOME will be searched.
+  -r,--rustup_home RUSTUP_HOME
+                        The rustup home. If not specify, system variable RUSTUP_HOME will be searched.
+  -v,--verbose          Verbose output.
+  -s,--strip_path       Strip rust file path.
+```
+
+example:<br>
+Command used under x86_64-unknown-uefi target:
+```
+cargo run -p td-reproducible-tool -- -n rust-td-payload --target x86_64-unknown-uefi -p release -v
+```
+is equal to 
+```
+cargo run -p td-reproducible-tool -- -w "." -n rust-td-payload.efi -t x86_64-unknown-uefi -p release -v
+```
+<br>
+Command used under x86_64-unknown-none target:
+
+```
+cargo run -p td-reproducible-tool -- -n rust-td-payload --target x86_64-unknown-none -p release -v
+```
--- a/td-reproducible-tool/src/main.rs
+++ b/td-reproducible-tool/src/main.rs
@@ -1,0 +1,214 @@
+// Copyright (c) 2022 Intel Corporation
+//
+// SPDX-License-Identifier: BSD-2-Clause-Patent
+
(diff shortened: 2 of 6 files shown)


gaojiaqi7:
The CI failure caused by secure boot verification is due to `strip = "symbols"`, there may be some hidden optimizations which are not compatible with the new setting

td-reproducible-tool/readme.md
```diff
@@ -0,0 +1,61 @@
...
+```
+1. Debug related sections (PE + ELF) 
+2. Code base path like \home\USERNAME\rust-td-payload\xxx\xxx.rs (PE + ELF) 
+3. CARGO_HOME like \home\USERNAME\.cargo\xxx\xxx.rs (PE + ELF) 
+4. RUSTUP_HOME like \home\USERNAME\.rustup\xxx\xxx.rs (PE + ELF) 
+5. TimeDateStamp field in COFF file header (PE only) 
+```
+
+Solution: <br>
+
```
jiangliu, line 16: Could 2, 3, 4 also be solved by using build container to normalize the build env?

jyao1: It is based upon the definition of "reproducible build".

In our definition, "reproducible build" means
1) Build with same code and version
2) Build with same compiler and version
3) Build at different time.
4) Build at different directory/path <==

td-reproducible-tool/readme.md
```diff
@@ -0,0 +1,61 @@
...
+```
+1. Use strip = "symbols" option to [profile.release]
```
jyao1, line 18: Please update to debuginfo

td-reproducible-tool/src/main.rs
```diff
@@ -0,0 +1,200 @@
...
+    .iter()
+    .collect();
+
+    assert!(binary_path.exists());
+    println!("INFO: Found the compiled file: {:?}", binary_path);
+
+    // check out CARGO_HOME and RUSTUP_HOME
+    cargo_home = if cargo_home == "" {
+        if env::var("CARGO_HOME").is_err() {
+            panic!("Neither --cargo_home nor system environment for \"CARGO_HOME\" is found! ")
```
jyao1, line 86: Can we make it optional?

jyao1, line 86: Can we make it optional?

td-reproducible-tool/src/main.rs
```diff
@@ -0,0 +1,200 @@
...
+        } else {
+            env::var("CARGO_HOME").unwrap()
+        }
+    } else {
+        cargo_home
+    };
+
+    rustup_home = if rustup_home == "" {
+        if env::var("RUSTUP_HOME").is_err() {
+            panic!("Neither --rustup_home nor system environment for \"RUSTUP_HOME\" is found! ")
```
jyao1, line 96: Can we make it optional?

jyao1, line 96: Can we make it optional?

longlongyang: It is optional now, please check the latest code change.

longlongyang: > Can we make it optional?

It is optional now, please check the latest code change.

td-reproducible-tool/src/main.rs
longlongyang: > Can we make it optional?

It is optional now, please check the latest code change.

longlongyang: > Can we make it optional?

It is optional now, please check the latest code change.

td-reproducible-tool/readme.md
longlongyang: > Please update to debuginfo

Still using symbols in the latest code change.

td-reproducible-tool/src/main.rs
```diff
@@ -0,0 +1,214 @@
...
+        } else {
+            env::var("CARGO_HOME").unwrap()
+        }
+    } else {
+        cargo_home
+    };
+
+    rustup_home = if rustup_home == "" {
+        if env::var("RUSTUP_HOME").is_err() {
+            panic!("Neither --rustup_home nor system environment for \"RUSTUP_HOME\" is found!\n")
```
jyao1, line 159: still panic?

td-reproducible-tool/src/main.rs
```diff
@@ -0,0 +1,214 @@
...
+        println!(
+            "INFO: -s or --strip_path is not specified, Skipping strip related rust file path."
+        );
+        return Ok(());
+    }
+
+    // Check out CARGO_HOME and RUSTUP_HOME, proceed to strip rust file path.
+    cargo_home = if cargo_home == "" {
+        if env::var("CARGO_HOME").is_err() {
+            panic!("Neither --cargo_home nor system environment for \"CARGO_HOME\" is found!\n")
```
jyao1, line 149: still panic?

td-reproducible-tool/readme.md
jyao1: The final design: 
1 and 2 are handled by build option.
3 and 4 are handled by post-build tool as an option.
5. Timestamp is always cleared.

td-reproducible-tool/src/main.rs
jyao1: OK. I did not found -s or --strip_path in readme.

Please add it.

longlongyang: > OK. I did not found -s or --strip_path in readme.
> 
> Please add it.

Sorry not to update readme. Now added, please check the latest readme.md

liuw1:
I validated reproducible build with the latest code, there are still some differences. And the secure boot issue was found in the latest CI test.

image

jyao1:
@liuw1 , please file a new issue, or reopen the old issue.

longlongyang:
> I validated reproducible build with the latest code, there are still some differences. And the secure boot issue was found in the latest CI test.
> 
> image

This position is the where TimeDateStamp sit at, Did you run reproducible tool after every binary built?
And are you sure you observed the secure boot issue using the commit newer or equal to commit#fc436aabae?

jyao1:
@longlongyang , I think we need update readme.md to add step for reproducible tool.
Otherwise, people may forget that step.

longlongyang:
> @longlongyang , I think we need update readme.md to add step for reproducible tool. Otherwise, people may forget that step.

Okay, will update.

PR #270, opened 2022-05-16
Update stable rust toolchain to 1.60.0

### LINKED ISSUE #269: Need to upgrade the toolchain to 1.59 release to accept "strip" option in release profile
Background: https://github.com/confidential-containers/td-shim/issues/256#issuecomment-1124460655
The current stable toolchain version 1.58 release doesn't support the "strip" option in release profile, which breaks the reproducible build. 
We need to upgrade the version of toolchain to 1.59 release, since the profile strip option has been stabilized in the 1.59 release. More detail please find: https://doc.rust-lang.org/cargo/reference/unstable.html#profile-strip-option

liuw1:
Fix #269

### DIFF
--- a/.github/workflows/devtools.yml
+++ b/.github/workflows/devtools.yml
@@ -13,7 +13,7 @@ env:
   AS: nasm
   AR_x86_64_unknown_uefi: llvm-ar
   CC_x86_64_unknown_uefi: clang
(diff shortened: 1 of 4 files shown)


jyao1 (approved):

jyao1:
Merge it to unblock the reproducible build issue.

PR #276, opened 2022-05-20
td-paging: make page table base and size as variable

gaojiaqi7:
Caller may decide where to put the page table and how much memory should page table occupy. This makes the td-paging crate more reusable.        
     - init() API will accept two arguments of base and size.
     - FrameAlloc use vector to hold the bitmap. Heap is ready when setting up page table.

### DIFF
--- a/td-layout/src/runtime.rs
+++ b/td-layout/src/runtime.rs
@@ -12,7 +12,7 @@
                     +--------------+ <-  0x00100000 (1M)
                     |   ........   |
                     +--------------+ <-  0x00800000
-                    |  Page Table  |
+                    |  Page Table  | <-  0x00800000
                     +--------------+ <-  0x01000000
                     |    TD HOB    |
                     +--------------+ <-  0x01100000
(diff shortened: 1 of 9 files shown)


td-layout/src/runtime.rs
```diff
@@ -44,6 +44,7 @@ pub const TD_PAYLOAD_SHADOW_STACK_SIZE: u32 = 0x200000;
 pub const TD_PAYLOAD_STACK_SIZE: u32 = 0x800000;
 
 pub const TD_PAYLOAD_PAGE_TABLE_BASE: u64 = 0x800000;
+pub const TD_PAGE_TABLE_SIZE: usize = 0x800000;
```
jiangliu, line 47: Should it be named as "TD_PAYLOAD_PAGE_TABLE_SIZE"?

gaojiaqi7: Yes it should be. I'll update the patch, thanks

jyao1 (approved):

PR #286, opened 2022-05-27
 Enhance tdx-tdcall crate

gaojiaqi7:
- To handle the case of many input/output operands, TdcallArgs and
   TdVmcallArgs are defined to gather all of the input and output operands
   to a structure and pass the pointer to the assembly function. The
   operands can be read and written inside the assembly function according
   to the memory offset.

 - Add two wrapper functions td_call and td_vmcall to ensure the correct
   use of the assembly function (i.e., asm_td_call and asm_td_vmcall).

 - When TDCALL/TDVMCALL returns error status code, functions other than
   instruction emulation (TDG.VP.VMCALL-Instruction-execution sub-functions)
   return an error instead of halt.


tdx-tdcall/src/asm/mod.rs
```diff
@@ -14,3 +15,8 @@ global_asm!(include_str!("tdcall.asm"));
 
 #[cfg(not(feature = "use_tdx_emulation"))]
 global_asm!(include_str!("tdvmcall.asm"));
+
+extern "win64" {
```
jyao1, line 19: Do we really need expose those APIs?

I am worried to expose unsafe FFI APIs.

jyao1:
please result conflict

jyao1 (approved):

PR #287, opened 2022-05-28
Add trusted boot guideline.

jyao1:

### DIFF
--- a/doc/tdshim_spec.md
+++ b/doc/tdshim_spec.md
@@ -824,11 +824,11 @@ in general.
 
 **Table 3.5-4 TD_SHIM_PLATFORM_CONFIG_INFO**
 
-| Field              | Byte Length | Byte Offset | Description                                 |
-|:-------------------|:------------|:------------|:--------------------------------------------|
-| Descriptor         | 16          | 0           | 16 bytes information specific descriptor.   |
-| InfoLength         | 4           | 16          | Length, in bytes, of the information field. |
-| Info               | InfoLength  | 20          | Information field.                          |
+| Type             | Name       | Description                                 |
+|:-----------------|:-----------|:--------------------------------------------|
+| BYTE[16]         | Descriptor | 16 bytes information specific descriptor.   |
+| UINT32           | InfoLength | Length, in bytes, of the information field. |
+| BYTE[InfoLength] | Info       | Information field.                          |
 
 ### Secure Boot Support
 
--- a/doc/tdshim_spec.md
+++ b/doc/tdshim_spec.md
@@ -762,6 +762,74 @@ The expected flow is:
    TDH.MR.EXTEND as described above, from the lowest GPA to the highest GPA,
    with 256 bytes granularity.
 
+#### Guideline
+
+TD Shim should follow [TCG Platform Firmware Profile specification](#tcg-pfp)
+in general.
+
+1. About separator
+
+   TD Shim shall measure 0x00000000 or 0xFFFFFFFF to RTMR[0~1] and the matching
+   EV_SEPARATOR event shall be recorded in the event log, prior to the
+   invocation of TD payload. As such, the verifier can know the rest events are
+   for the TD payload.
+
+   In case of an error occurs, TD shim shall measure 0x00000001 to RTMR[0~1] and
+   the matching EV_SEPARATOR event shall be recorded in the event log.
+
+2. About TD configuration
+
+   TD Shim shall extend the TD Hob to RTMR[0] and the event should be
+   EV_PLATFORM_CONFIG_FLAGS with event data being TD_SHIM_PLATFORM_CONFIG_INFO
+   and Descriptor being "td_hob\0", Info being the whole TD Hob.
+
+3. About payload info
+
+   TD Shim shall extend the payload binary to RTMR[1] and the event should be
+   EV_EFI_PLATFORM_FIRMWARE_BLOB2 with event data being
+   UEFI_PLATFORM_FIRMWARE_BLOB2 and BlobDescription being "td_payload\0".
+
+   TD Shim shall extend the payload parameter to RTMR[1] and the event should be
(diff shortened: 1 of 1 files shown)


doc/tdshim_spec.md
```diff
@@ -762,6 +762,55 @@ The expected flow is:
...
+
+3. About payload info
+
+   TD Shim shall extend the payload binary to RTMR[1] and the event should be
+   EV_EFI_PLATFORM_FIRMWARE_BLOB2 with event data being
+   UEFI_PLATFORM_FIRMWARE_BLOB2 and BlobDescription being "td_payload".
+
+   TD Shim shall extend the payload parameter to RTMR[1] and the event should be
+   EV_EFI_PLATFORM_FIRMWARE_BLOB2 with event data being
+   UEFI_PLATFORM_FIRMWARE_BLOB2 and BlobDescription being "td_payload_parameter".
```
jialez0, line 794: Do we need to include the content of the payload parameter in the event data instead of just a descriptive information?

jyao1: Do you mean "td_payload_parameter:\<param_string\>" ?

According to TCG, the size of descriptor is 256 byte. I am not sure if it is big enough..
Is truncation in descriptor acceptable?

jialez0: What do you mean by truncation? Does this mean that if one event entry is not enough, use multiple entries to record the contents of the payload parameter?

jyao1: Sorry for the confusion. Truncation means only partial of the data is recorded. For example:
Let's assume the descriptor space is small, only allows 24 bytes.
While the boot parameter is very long "1234567890abcdef1234567890abcdef" - 32 bytes.
The hash will be HASH("1234567890abcdef1234567890abcdef") - no truncation.
But the descriptor will be "1234567890abcdef12345678" - it is truncated.

I am thinking below options:

A. Current UEFI_PLATFORM_FIRMWARE_BLOB2 includes a pointer and length for the data. But not data itself. that is why I thought to put parameter string in descriptor.
```
typedef struct tdUEFI_PLATFORM_FIRMWARE_BLOB2 {
    UINT8 BlobDescriptionSize;
    BYTE[BlobDescriptionSize] BlobDescription;
    UEFI_PHYSICAL_ADDRESS BlobBase;
    UINT64 BlobLength;
} UEFI_PLATFORM_FIRMWARE_BLOB2;
```

If you can put parameter string in td-shim image, then the verifier can already find the data by using pointer. No need to put parameter string in descriptor.

The flow is:
1. Verifier finds the "td_payload_parameter"
2. Verifier gets the UEFI_PLATFORM_FIRMWARE_BLOB2.BlobBase and BlobLength.
3. Verifier gets the corresponding parameter string there.
Does that resolve the problem?


If you really really want to include the data, then we have to use other type. For example:

B. EV_PLATFORM_CONFIG_FLAGS.
The event data could be the parameter string.

What do you think?

jialez0: I think scheme B may be more appropriate, because in normal verification scenarios, the verifier may not be able to get and load the td-shim image. Using scheme B allows the verifier to verify the parameter settings in a more flexible and lightweight way.

jialez0: I think scheme B may be more appropriate because if the verifier obtains and loads the referenced TD shim image and reads the parameters through the pointer in event data, if the parameters measured in the eventlog have been tampered with, the verifier cannot accurately know which parameters are not configured as expected.

jyao1: agree. Will update the proposal.

jyao1:
@jialez0 , would you please review the new proposal?

doc/tdshim_spec.md
```diff
@@ -762,6 +762,74 @@ The expected flow is:
...
+
+3. About payload info
+
+   TD Shim shall extend the payload binary to RTMR[1] and the event should be
+   EV_EFI_PLATFORM_FIRMWARE_BLOB2 with event data being
+   UEFI_PLATFORM_FIRMWARE_BLOB2 and BlobDescription being "td_payload\0".
+
+   TD Shim shall extend the payload parameter to RTMR[1] and the event should be
+   EV_PLATFORM_CONFIG_FLAGS with event data being TD_SHIM_PLATFORM_CONFIG_INFO
+   and Descriptor being "td_payload_info\0", Info being the whole TD Hob.
```
jialez0, line 794: `Info being the whole TD Hob` -> `Info being the whole payload parameters`. Right?

jialez0 (approved):

doc/tdshim_spec.md
```diff
@@ -762,6 +762,74 @@ The expected flow is:
...
+
+   TD Shim shall include the TCG_EfiSpecIDEvent as the first event and the
+   vendorInfo should be "td_shim".
+
+   TD Shim may include TCG_Sp800_155_PlatformId_Event2 to provide reference
+   manifest information.
+
+**Table 3.5-4 TD_SHIM_PLATFORM_CONFIG_INFO**
+
+| Field              | Byte Length | Byte Offset | Description                                 |
```
jodh-intel, line 827: Nit: It would be slightly clearer (atleast to me) if the "Byte Offset" column came before the "Byte Length" column.

jyao1: I notice that different standards are using different style, for example:
* UEFI/UEFI is using Mnemonic/Offset/Length/Description.
* UEFI/ACPI is using Field/Length/Offset/Description.
* TCG/PFP is using Type/Name/Description.
* TCG/MOR is using Offset/Size/Name/Description.
* DMTF/SPDM is using Offset/Field/Size/Description
* DMTF/SMBIOS is using Offset/Version/Name/Length/Value/Description

I will choose the style similar to existing standard.
Since we are defining something similar to TCG/PFP. I will use - Type/Name/Description.

Issue #272, opened 2022-05-18
[Bug] secure_boot_verify_payload fails with InvalidSignature when using ECDSA_NIST_P384_SHA384 crypto suite


longlongyang:
secure_boot_verify_payload fails with InvalidSignature when using ECDSA_NIST_P384_SHA384 crypto suite with strip = "symbols" set in cargo.toml.
Location:
https://github.com/confidential-containers/td-shim/blob/552e40dd0b5b3dadc3bd9ae30c7b223d26cfe7f6/td-shim/src/bin/td-shim/verifier.rs#L162-L164
How to reproduce the bug:

- If no panic is observed using the code change https://github.com/confidential-containers/td-shim/pull/268, 

  Try add 
  `log::info!("Just visit the image memory, and the bug will be triggered: {:?}\n", self.image[100]);`
  to the line before line 162
  https://github.com/confidential-containers/td-shim/blob/552e40dd0b5b3dadc3bd9ae30c7b223d26cfe7f6/td-shim/src/bin/td-shim/verifier.rs#L162

- if panic is observed using the code change https://github.com/confidential-containers/td-shim/pull/268, then you have reproduced the bug.

Note: The trigger condition of this bug is not fixed. At very first, I caught the bug without visiting the image(message) memory, and the bug is gone after have a visiting in the image memory; And in the other day, I have to visit the image memory in order to trigger the bug.

FYR:
The signer:
https://github.com/confidential-containers/td-shim/blob/552e40dd0b5b3dadc3bd9ae30c7b223d26cfe7f6/td-shim-tools/src/signer.rs#L103-L122
The verifier:
https://github.com/confidential-containers/td-shim/blob/552e40dd0b5b3dadc3bd9ae30c7b223d26cfe7f6/td-shim/src/bin/td-shim/verifier.rs#L55-L89

longlongyang:
may be caused by https://github.com/confidential-containers/td-shim/issues/279

ariel-adam:
@longlongyang is this issue still relevant or can be closed? 
If it's still relevant to what release do you think we should map it to (mid-November, end-December, mid-February etc...)?

ariel-adam:
@longlongyang is this issue still relevant or can be closed? 
If it's still relevant to what release do you think we should map it to (mid-November, end-December, mid-February etc...)?

longlongyang:
[...]

PR #296, opened 2022-06-01
reproducible-tool: Roll back temporarily due to bug introduced by strip option

longlongyang:

### DIFF
--- a/td-reproducible-tool/readme.md
+++ b/td-reproducible-tool/readme.md
@@ -23,6 +23,8 @@ Solution: <br>
 ```
 NOTE: 3 and 4 may be resolved directly after rust-lang RFC 3127. This tool provide a temp solution to solve 3 and 4 before RFC is implemented.
 
+NOTE: 1 and 2 are not applied because this strip feature has potential stability issue. For example, [bug](https://github.com/confidential-containers/td-shim/issues/272). It can be enabled after the strip becomes more robust.
+
 ## tool usage
 
 ```
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,4 +28,3 @@ panic = "abort" # disable stack unwinding on panic
 [profile.release]
 panic = "abort"    # disable stack unwinding on panic
 lto = true         # Link-time optimization
-strip = "symbols"  # remove debug related symbols/sections


td-reproducible-tool/readme.md
```diff
@@ -23,6 +23,8 @@ Solution: <br>
 ```
 NOTE: 3 and 4 may be resolved directly after rust-lang RFC 3127. This tool provide a temp solution to solve 3 and 4 before RFC is implemented.
 
+NOTE: Since there is a [bug](https://github.com/confidential-containers/td-shim/issues/272) that maybe was introduced by the strip option, So now rolling back temporarily.
```
jyao1, line 26: I recommend use below:

NOTE: 1 and 2 are not applied because this strip feature has potential stability issue. For example, [bug](https://github.com/confidential-containers/td-shim/issues/272). It can be enabled after the strip becomes more robust.

td-reproducible-tool/src/main.rs
```diff
@@ -140,6 +140,7 @@ fn main() -> std::io::Result<()> {
         println!(
             "INFO: -s or --strip_path is not specified, Skipping strip related rust file path."
         );
+        fs::write(binary_path, buf).expect("Failed to open the compiled binary!\n");
```
jyao1, line 143: This should be a bug fix, not related to rollback.

Please create another patch for this specific issue. Dont mix them.

td-reproducible-tool/readme.md
longlongyang: > I recommend use below:
> 
> NOTE: 1 and 2 are not applied because this strip feature has potential stability issue. For example, [bug](https://github.com/confidential-containers/td-shim/issues/272). It can be enabled after the strip becomes more robust.

Adopted, thank you for your advice!

td-reproducible-tool/src/main.rs
longlongyang: > This should be a bug fix, not related to rollback.
> 
> Please create another patch for this specific issue. Dont mix them.

Ok, Got it.

jyao1 (approved):

jyao1:
Urgent fix for CI. Merge.

PR #303, opened 2022-06-05
reproducible-tool: clear timedatestamp and guid in PE image

### LINKED ISSUE #302: need clear more debug info in PE image
1) The `TimeDateStamp` in debug directory entry is not cleared.

REF: https://docs.microsoft.com/en-us/windows/win32/debug/pe-format

2) The `Guid` in CodeView Debug Directory Entry is not cleared

REF: https://github.com/dotnet/runtime/blob/main/docs/design/specs/PE-COFF.md

Overall, please refer to `ZeroDebugData()` in https://github.com/tianocore/edk2/blob/master/BaseTools/Source/C/GenFw/GenFw.c

longlongyang:
fix #302

### DIFF
--- a/td-shim-tools/src/bin/td-shim-strip-info/main.rs
+++ b/td-shim-tools/src/bin/td-shim-strip-info/main.rs
@@ -9,12 +9,108 @@ use regex::bytes::Regex;
 use std::convert::TryInto;
 use std::fs;
 use std::io::Read;
+use std::mem::size_of;
 use std::{env, fs::File, path::PathBuf};
 use zeroize::Zeroize;
 
 // https://docs.microsoft.com/en-us/windows/win32/debug/pe-format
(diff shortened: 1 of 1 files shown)


td-reproducible-tool/Cargo.toml
```diff
@@ -15,4 +15,5 @@ default = []
 [dependencies]
 argparse = "0.2.2"
 zeroize = "1.5.4"
-regex = "1"
\ No newline at end of file
+regex = "1"
+codec = { git="https://github.com/jyao1/rust-spdm", default-features = false }
```
jyao1, line 19: Please dont depend on this.

longlongyang: > Please dont depend on this.

Dependency is now removed in latest code change.

PR #304, opened 2022-06-07
td-shim: follow the td-shim trust boot guideline

gaojiaqi7:
- Record the TCG_EfiSpecIDEvent as the first event.

 - For TD HOB, extend it to RTMR[0] and record the event with event type EV_PLATFORM_CONFIG_FLAGS and event data being TD_SHIM_PLATFORM_CONFIG_INFO.

 - For payload binary, extend it to RTMR[1] and record the event with event type EV_EFI_PLATFORM_FIRMWARE_BLOB2 and event data being UEFI_PLATFORM_FIRMWARE_BLOB2.

 - For secure boot, extend the provisioned and matched trust anchor into RTMR[0] and payload SVN into RTMR[1] and record them with event type EV_PLATFORM_CONFIG_FLAGS and event data being TD_SHIM_PLATFORM_CONFIG_INFO.

 - Before jumping to the TD payload, td-shim measures 0x0000_0000 and record the EV_SEPARATOR into the event log to mark the end of the td-shim events.

### DIFF
--- a/td-shim/src/bin/td-shim/tcg.rs
+++ b/td-shim/src/bin/td-shim/tcg.rs
@@ -3,11 +3,19 @@
 // SPDX-License-Identifier: BSD-2-Clause-Patent
 
 use core::convert::TryInto;
+use core::mem::size_of;
 use scroll::{Pread, Pwrite};
 use td_shim::event_log::{
-    CcEventHeader, Ccel, TpmlDigestValues, TpmtHa, TpmuHa, CCEL_CC_TYPE_TDX, CC_EVENT_HEADER_SIZE,
-    SHA384_DIGEST_SIZE, TPML_ALG_SHA384,
+    CcEventHeader, Ccel, TcgEfiSpecIdevent, TdShimPlatformConfigInfoHeader, TpmlDigestValues,
+    TpmtHa, TpmuHa, CCEL_CC_TYPE_TDX, CC_EVENT_HEADER_SIZE, EV_NO_ACTION, EV_PLATFORM_CONFIG_FLAGS,
+    EV_SEPARATOR, SHA384_DIGEST_SIZE, TPML_ALG_SHA384,
 };
+
+#[derive(Debug)]
+pub enum TdEventLogError {
+    OutOfResource,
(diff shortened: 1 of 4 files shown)


td-shim/src/bin/td-shim/tcg.rs
```diff
@@ -34,7 +42,20 @@ impl TdEventLog {
...
+        };
+
+        // Create the TCG_EfiSpecIDEvent as the first event
+        let first = TcgEfiSpecIdevent::default();
+        td_event_log
+            .log_event(
+                0,
+                EV_NO_ACTION,
+                first.as_bytes(),
+                &[0u8; SHA384_DIGEST_SIZE],
```
jyao1, line 54: The digest for SpecidEvent is always 20 bytes - to align with SHA1 for compatibility reason.

td-shim/src/event_log.rs
```diff
@@ -16,10 +19,143 @@ pub const PCR_DIGEST_NUM: usize = 1;
...
+
+    pub fn as_bytes(&self) -> &[u8] {
+        unsafe { &*slice_from_raw_parts(self as *const Self as *const u8, size_of::<Self>()) }
+    }
+}
+
+impl Default for TcgEfiSpecIdevent {
+    fn default() -> Self {
+        Self {
+            signature: *b"Spec ID Event01\0",
```
jyao1, line 79: Why you use "01", the PFP spec says "03", refer to Table 20 TCG_EfiSpecIdEvent.

td-shim/src/event_log.rs
```diff
@@ -16,10 +19,143 @@ pub const PCR_DIGEST_NUM: usize = 1;
...
+            platform_class: 0,
+            spec_version_minor: 0,
+            spec_version_major: 0x2,
+            spec_errata: 0,
```
jyao1, line 83: errata is 105

td-shim/src/event_log.rs
gaojiaqi7: Yes it should be '03', misunderstood the content and thought it is an example

PR #307, opened 2022-06-12
tdx-tdcall: correct the expose register mask 

gaojiaqi7:


tdx-tdcall/src/asm/tdvmcall.asm
```diff
@@ -8,7 +8,7 @@
 # corresponding register is passed to VMM. Refer to TDX Module
 # ABI specification section TDG.VP.VMCALL for detail.
 # Here we expose R10 - R15 to VMM in td_vm_call()
-.equ TDVMCALL_EXPOSE_REGS_MASK, 0xffec
+.equ TDVMCALL_EXPOSE_REGS_MASK, 0xfc00
```
jyao1, line 11: Why???
According to TDX spec: only RAX,RCX,RSP must be 0. Other should be 1.

gaojiaqi7: because no other registers are used, it is useless to expose these registers to VMM. Spec says RAX,RCX,RSP and reserved must be 0, others can be controlled by TD

jyao1: The GHCI may add more interfaces. Other components may define extra TDVMCALL.

The generic TDX API should only follow TDX spec.

gaojiaqi7: Right, but implementation in asm_td_vmcall() only passes values of r10-r15 to VMM, if there's update in GHCI, we need to update our implementation both with the 'mask' value.
It seems redundant to zero out other registers and expose them to VMM .

jyao1: I think the direction is:
1) tdvmcall generic interface should follow TDX-module spec. (including all GPR mask, XMM mask). The goal is to ensure no tdvmcall API change, if a new GHCI API requires more GPRs or XMMs.

2) tdvmcall individual interface should follow GHCI spec. If we add more registers in current API in GHCI, we may change the API. If we add new API in GHCI, we may add new API.

One possible option is to pass the mask as parameter.

jyao1 (approved):

PR #312, opened 2022-06-16
doc: update td-shim spec to support lazy-accept

gaojiaqi7:
Fix: #315

According to the ongoing patches of guest Kernel:
https://github.com/intel/tdx/tree/guest-unaccepted-memory, it does not
support the unaccepted type in E820 table and uses a bitmap to track the
unaccepted memory pages instead.

Since td-shim does not provide UEFI interface for OS to obtain the memory
map, it shall build a bitmap as it does in kernel EFI stub and
report the bitmap address to the Kernel through boot paramter.

The bitmap memory does not need to be explicitly reserved, because the
kernel will reserve it before initializing the memblock.

### DIFF
--- a/doc/tdshim_spec.md
+++ b/doc/tdshim_spec.md
@@ -625,8 +625,17 @@ If a TD-memory region is shared memory, the TD Shim shall convert it to private
 memory before transfer to a payload.
 
 If a TD-memory region is unaccepted memory and requires TDCALL
-[TDG.MEM.PAGE.ACCEPT] in the TD guest OS, then the TD Shim shall have the final
-memory map report this region with AddressRangeUnaccepted.
+[TDG.MEM.PAGE.ACCEPT] in the guest TD payload, then the TD Shim shall mark this
+region as unaccepted momory to payload. For the payload supporting Linux Boot
+Protocol, the TD Shim shall have the final memory map report this region
+with AddressRangeMemory and build a bitmap for all the system memory to mark
+whether each 2MB size page has been accepted or not. A bit set to 1 in
+bitmap indicates that the corresponding page has not been accepted yet. The
+TD Shim shall write the 64-bit bitmap address to boot parameter
+([offset 0x78](https://github.com/intel/tdx/blob/guest-unaccepted-memory/Documentation/x86/zero-page.rst)).
+and report the bitmap memory as AddressRangeMemory in E820 table. For the
+payload not supporting Linux Boot Protocol, the TD Shim shall have the final
+memory map report unaccepted memory region with AddressRangeUnaccepted.
 
 If a memory region is MMIO, it is designed to only be accessed via
 TDVMCALL<#VE.RequestMMIO> and not via direct memory read or write.  Accordingly,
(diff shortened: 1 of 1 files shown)


doc/tdshim_spec.md
```diff
@@ -626,7 +626,7 @@ memory before transfer to a payload.
 
 If a TD-memory region is unaccepted memory and requires TDCALL
 [TDG.MEM.PAGE.ACCEPT] in the TD guest OS, then the TD Shim shall have the final
-memory map report this region with AddressRangeUnaccepted.
+memory map report this region with AddressRangeMemory and build a bitmap for all the system memory to mark whether each 2MB size page has been accepted or not. A bit set to 1 in bitmap indicates that the corresponding page has not been accepted yet. For the payload supporting Linux Boot Protocol, the TD Shim shall write the 64-bit bitmap address to boot parameter (offset 0x78).
```
jyao1, line 629: please add link for the offset 0x78.

jyao1, line 629: please make sure each line is 80 char.

doc/tdshim_spec.md
```diff
@@ -640,11 +645,10 @@ Table 3.4-1 shows the E820 memory map.
 
 | E820 Memory Type       | Value | Usage                                                                 | TD-Memory Type | OS Action                          |
 |:-----------------------|:------|:----------------------------------------------------------------------|:---------------|:-----------------------------------|
-| AddressRangeMemory     | 1     | Usable by OS.                                                         | Private        | Use directly                       |
+| AddressRangeMemory     | 1     | Usable by OS.                                                         | Private/Unaccepted | Use directly                       |
```
jyao1, line 648: please use `Private (Accepted or Unaccepted)`.

OS Action `For Accepted, use directly. For Unaccepted, use after convert to private page.`

doc/tdshim_spec.md
```diff
@@ -640,11 +645,10 @@ Table 3.4-1 shows the E820 memory map.
...
 | AddressRangeReserved   | 2     | Firmware-Reserved region, such as flash.                              | Private        | Reserved                           |
 | AddressRangeACPI       | 3     | ACPI table.                                                           | Private        | Use after copy ACPI table.         |
 | AddressRangeNVS        | 4     | Firmware Reserved for ACPI, such as the memory used in ACPI OpRegion. | Private        | Reserved                           |
-| AddressRangeUnaccepted | 8     | Allocated by VMM, but not accepted by TD guest yet.                   | Unaccepted     | Use after convert to private page. |
```
jyao1, line 647: I think we still need this for non-Linux Boot Protocol, such as payload.

doc/tdshim_spec.md
```diff
@@ -626,7 +626,12 @@ memory before transfer to a payload.
 
 If a TD-memory region is unaccepted memory and requires TDCALL
 [TDG.MEM.PAGE.ACCEPT] in the TD guest OS, then the TD Shim shall have the final
-memory map report this region with AddressRangeUnaccepted.
+memory map report this region with AddressRangeMemory and build a bitmap for
+all the system memory to mark whether each 2MB size page has been accepted or
+not. A bit set to 1 in bitmap indicates that the corresponding page has not
+been accepted yet. For the payload supporting Linux Boot Protocol, the TD Shim
+shall write the 64-bit bitmap address to boot parameter
+([offset 0x78](https://github.com/intel/tdx/blob/guest-unaccepted-memory/Documentation/x86/zero-page.rst)).
```
jyao1, line 634: Please describe the memory type of the bitmap as well. (e.g. normal memory, or reserved)

gaojiaqi7: Sure

jyao1:
please also add a small section to describe the "optimization" concept under "Memory Map" section.

#### Optimization Consideration
Usually, td-shim need add all physical pages to private memory before transfer control to the payload. It may increase the boot time, because TDCALL[TDG.MEM.PAGE.ACCEPT] is time-consuming.

The td-shim may choose to add part of memory to be private memory and let payload convert the rest pages. This is called lazy-accept. For example, td-shim may accept the first 256MiB, and mark the reset pages to be unaccepted in E820 table and boot_parameter.

jyao1 (approved):

PR #313, opened 2022-06-16
Introduce lazy-accept to improve boot performance

gaojiaqi7:
Lazy-accept means only part of the unaccepted memory is accepted by the td-shim, and the rest is handled by the OS kernel.

Since td-shim does not provide UEFI interface for OS to obtain the memory map, it should build a bitmap to mark whether a 2MB page has been accepted or not, and pass the bitmap address to OS, e.g., the boot parameter (offset 0x78) for Linux Kernel.

To make this feature configurable:
 - Add unaccepted memory bitmap in runtime layout, it is used to save a bitmap that marks whether a 2MB page has been accepted by td-shim or not. The default 256KB bitmap can track 4TB system memory.
 - Partial accept memory size in runtime layout config is used to indicate how much memory should be accepted when lazy accept is enabled. The default value is 256MB.
 - Feature 'lazy-accept' is introduced in td-shim crate to enable the lazy accept function.

### DIFF
--- a/td-shim/Cargo.toml
+++ b/td-shim/Cargo.toml
@@ -50,6 +50,7 @@ boot-kernel = ["td-layout/boot-kernel"]
 cet-ss = []
 secure-boot = ["der"]
 tdx = ["tdx-tdcall", "td-exception/tdx", "td-logger/tdx", "x86"]
+lazy-accept = ["tdx"]
 main = [
     "td-loader",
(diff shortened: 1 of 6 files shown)


jyao1:
Do you have any data on perf improvement?
E.g. if we have 4G guest memory, what is before? what is after?

PR #314, opened 2022-06-17
tests: counting memory size ignores the reserved address range

gaojiaqi7:
Address range from top of low memory (TOLM) to 4G is not for system memory
so we should exclude this range when calculating the system memory size.

### DIFF
--- a/tests/test-td-payload/src/testmemmap.rs
+++ b/tests/test-td-payload/src/testmemmap.rs
@@ -18,6 +18,9 @@ use td_shim::TD_E820_TABLE_HOB_GUID;
 use td_uefi_pi::hob;
 use zerocopy::{AsBytes, FromBytes};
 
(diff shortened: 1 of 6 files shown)


tests/test-td-payload/src/testmemmap.rs
```diff
@@ -102,6 +105,14 @@ impl TestMemoryMap {
                 log::error!("Invalid E820 entry: {:x?}\n", entry);
                 return TestResult::Fail;
             }
+
+            // Do not count the memory in reserved range into total memory size
```
jyao1, line 109: Please add detail description on why KVM add reserved pages for EPT.

jyao1 (changes requested):
Please add comment on why KVM add reserved pages.

jyao1 (approved):

PR #319, opened 2022-06-21
Add check for readlen and header.length in secure_boot.rs

### LINKED ISSUE #317: [BUG] Td-shim will panic if header.length less than readlen
Line 278 in td-shim/src/secure_boot.rs
let trusted_hash = &file[readlen..header.length as usize];

```fuzz/target/x86_64-unknown-linux-gnu/release/secure_boot_cfv: Running 1 inputs 1 time(s) each.
Running: fuzz/artifacts/secure_boot_cfv/crash-8325968ba3195143553381b6182c8be6b809253a
thread '<unnamed>' panicked at 'slice index starts at 40 but ends at 0', library/core/src/slice/index.rs:92:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace```

liuw1:
Fix #317


jyao1 (approved):

jodh-intel (approved):
Thanks @liuw1.

It would be great if you could add a unit test for this.

PR #324, opened 2022-06-24
td-loader: Fix build issue of fuzzing test case

liuw1:
Remove the third parameter in relocate_elf_with_per_program_header()


jyao1 (approved):

liuw1:
> Thanks @liuw1.
> 
> lgtm
> 
> Do we need a CI test for the fuzzer to avoid future regressions?

I think we can add build test for fuzzer in CI to avoid future regression.

PR #334, opened 2022-07-06
Boot rawvmlinux

a74731248:
To boot RawVmlinux image.

Signed-off-by: Yonghao Wei <weiyonghao@linux.alibaba.com>


gaojiaqi7 (commented):
hi @a74731248 , have you tried how much memory may be used to load the kernel?

a74731248:
> hi @a74731248 , have you tried how much memory may be used to load the kernel?

I have tried several kernel images, and results show that how much memory is needed related with the kernel size and `0x8000000` is the one to satisfy all these images. I think the paylaod size can be flexibly configured.

jyao1 (approved):

Issue #340, opened 2022-07-19
Td-shim will panic in line 52 of td-uefi-pi/src/hob.rs


liuw1:
If phit.efi_end_of_hob_list less than hob.as_ptr(), Td-shim will panic.

PR #341, opened 2022-07-19
td-uefi-pi: Fix panic issue in line 52 of src/hob.rs

liuw1:
Fix: #340

### DIFF
--- a/td-uefi-pi/src/hob.rs
+++ b/td-uefi-pi/src/hob.rs
@@ -46,10 +46,11 @@ pub fn seek_to_next_hob(hob_list: &'_ [u8]) -> Option<&'_ [u8]> {
 // Check hob length equal efi_end_of_hob_list of HandoffInfoTable
 pub fn check_hob_length(hob: &[u8], hob_length: usize) -> Option<&[u8]> {
     let phit: HandoffInfoTable = hob.pread(0).ok()?;
+    let end: u64 = phit.efi_end_of_hob_list;
     if phit.header.r#type == HOB_TYPE_HANDOFF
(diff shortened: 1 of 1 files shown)


td-uefi-pi/src/hob.rs
```diff
@@ -49,7 +49,10 @@ pub fn check_hob_length(hob: &[u8], hob_length: usize) -> Option<&[u8]> {
     if phit.header.r#type == HOB_TYPE_HANDOFF
         && phit.header.length as usize >= size_of::<HandoffInfoTable>()
     {
-        if hob_length as u64 == phit.efi_end_of_hob_list - hob.as_ptr() as u64 {
+        if u64::MAX - (hob_length as u64) < hob.as_ptr() as u64 {
```
jyao1, line 52: Can we use rust checked_sub function to handle this instead of use u64::MAX?

liuw1: Replaced

td-uefi-pi/src/hob.rs
```diff
@@ -49,6 +49,7 @@ pub fn check_hob_length(hob: &[u8], hob_length: usize) -> Option<&[u8]> {
     if phit.header.r#type == HOB_TYPE_HANDOFF
         && phit.header.length as usize >= size_of::<HandoffInfoTable>()
     {
+        phit.efi_end_of_hob_list.checked_sub(hob.as_ptr() as u64)?;
```
jyao1, line 52: I think we just need sub once. No need to sub twice.

liuw1: Fixed

jyao1 (approved):

Issue #362, opened 2022-08-09
Assertion might fail with "attempt to add with overflow" in line 528 of elf64.rs


liuw1:
error: [Prusti: verification error] assertion might fail with "attempt to add with overflow"
   --> td-loader/src/elf64.rs:528:32
    |
528 |         self.sh_addr as usize..(self.sh_addr + self.sh_size) as usize
    |                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

PR #367, opened 2022-08-10
td-loader: Use checked_add to avoid arithmetic overflow

liuw1:
Fix: #362


jodh-intel (approved):
Thanks @liuw1.

lgtm, but could you also add a unit test for this change?

liuw1:
> Thanks @liuw1.
> 
> lgtm, but could you also add a unit test for this change?

Added

jodh-intel:
Thanks @liuw1/

jyao1 (approved):

Issue #370, opened 2022-08-10
Assertion might fail with "attempt to add with overflow" in line 94 of fv.rs


liuw1:
error: [Prusti: verification error] assertion might fail with "attempt to add with overflow"
  --> td-uefi-pi/src/pi/fv.rs:94:25
   |
94 |         self.checksum = u16::MAX - sum16(self.as_bytes()) + 1;
   |

PR #371, opened 2022-08-10
td-uefi-pi: Use wrapping_add to avoid overflow

liuw1:
Fix: #370
Fix: #375 


jodh-intel (approved):
Thanks @liuw1 - lgtm.

Again, could you add a unit test to prove that overflow cannot occur now?

liuw1:
> Thanks @liuw1 - lgtm.
> 
> Again, could you add a unit test to prove that overflow cannot occur now?

Added, and found another same type issue in line 262. Fix this issue in this PR.

jyao1 (approved):

Issue #364, opened 2022-08-09
Value might not fit into the target type in line 106 127 and 148 of tdx.rs


liuw1:
error: [Prusti: verification error] value might not fit into the target type.
   --> tdx-tdcall/src/tdx.rs:106:5
    |
106 |     args.r11 as u8
    |     ^^^^^^^^^^^^^^

error: [Prusti: verification error] value might not fit into the target type.
   --> tdx-tdcall/src/tdx.rs:127:5
    |
127 |     args.r11 as u16
error: [Prusti: verification error] value might not fit into the target type.
   --> tdx-tdcall/src/tdx.rs:148:5
    |
148 |     args.r11 as u32
    |     ^^^^^^^^^^^^^^^

error: [Prusti: verification error] value might not fit into the target type.
   --> tdx-tdcall/src/tdx.rs:348:14
    |
348 |         eax: args.r12 as u32,
    |              ^^^^^^^^^^^^^^^

liuw1:
After investigation, it is not an issue. We can close it.

PR #374, opened 2022-08-11
tdx-tdcall: Fix "Value might not fit into the target type" issue

liuw1:
Fix: #364

### DIFF
--- a/tdx-tdcall/src/tdx.rs
+++ b/tdx-tdcall/src/tdx.rs
@@ -103,7 +103,7 @@ pub fn tdvmcall_io_read_8(port: u16) -> u8 {
         tdvmcall_halt();
     }
 
-    args.r11 as u8
+    (args.r11 & 0xff) as u8
 }
 
 /// Request the VMM perform 2-bytes byte IO read operation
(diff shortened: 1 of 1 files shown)


tdx-tdcall/src/tdx.rs
```diff
@@ -345,10 +345,10 @@ pub fn tdvmcall_cpuid(eax: u32, ecx: u32) -> CpuIdInfo {
     }
 
     CpuIdInfo {
-        eax: args.r12 as u32,
-        ebx: args.r13 as u32,
-        ecx: args.r14 as u32,
-        edx: args.r15 as u32,
+        eax: (args.r12 & 0xffffffff) as u32,
```
jodh-intel, line 348: Nit: Might be worth using an underscore for clarity (here and elsewhere):

```suggestion
        eax: (args.r12 & 0xffff_ffff) as u32,
```

jodh-intel (approved):
Thanks @liuw1.

tdx-tdcall/src/tdx.rs
liuw1: Thanks! Fixed

jyao1 (approved):

PR #378, opened 2022-08-17
td-shim-tools: perform basic check for public key

gaojiaqi7:

### DIFF
--- a/td-shim-tools/src/enroller.rs
+++ b/td-shim-tools/src/enroller.rs
@@ -161,6 +161,17 @@ pub fn enroll_files(
     Ok(())
 }
 
+// Uncompressed ecdsa p384 public key length
+const ECDSA_P384_PUB_KEY_LEN: usize = 96;
+// Prefix of uncompressed ecdsa public key
+const ECDSA_UNCOMPRESSED_PUB_KEY_PREFIX: u8 = 0x04;
+// Minimal RSA exponent
+const RSA_EXPONENT: u64 = 0x10001;
+// RSA 3072 public key modulus length
+const RSA_3072_PUB_KEY_MOD_LEN: usize = 384;
+// Maximum size of public key file (1024 KiB)
+const PUB_KEY_MAX_SIZE: usize = 1024 * 1024;
+
 /// Build a firmware file which contains public key bytes for secure boot.
 ///
 /// Secure boot in td-shim means the td-shim will verify the digital signature of the payload,
@@ -180,7 +191,7 @@ pub fn create_key_file(key_file: &str, hash_alg: &str) -> io::Result<FirmwareRaw
         }
     };
 
-    let key_data = InputData::new(key_file, 1..=1024 * 1024, "public key")?;
+    let key_data = InputData::new(key_file, 1..=PUB_KEY_MAX_SIZE, "public key")?;
     let key = SubjectPublicKeyInfo::try_from(key_data.as_bytes()).map_err(|e| {
         error!("Can not load key from file {}: {}", key_file, e);
         io::Error::new(io::ErrorKind::Other, "invalid key data")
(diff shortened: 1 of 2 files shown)


td-shim-tools/src/enroller.rs
```diff
@@ -230,6 +241,17 @@ pub fn create_key_file(key_file: &str, hash_alg: &str) -> io::Result<FirmwareRaw
             }
             exp_bytes[8 - pubkey.exponents.as_bytes().len()..]
                 .copy_from_slice(pubkey.exponents.as_bytes());
+
+            let exp = u64::from_be_bytes(exp_bytes);
+
+            // According to FIPS 186-5, RSA exponent should be an odd and in range (2^16, 2^256)
+            if exp < RSA_MIN_EXPONENT
+                || exp % 2 == 0
```
jyao1, line 249: Please check with 0x10001

td-shim-tools/src/enroller.rs
```diff
@@ -161,6 +161,15 @@ pub fn enroll_files(
     Ok(())
 }
 
+// Uncompressed ecdsa p384 public key length
+const ECDSA_P384_PUB_KEY_LEN: usize = 97;
```
jyao1, line 165: Why it is 97?
The pub key (x + y) should be 48 + 48 = 96

td-shim-tools/src/enroller.rs
```diff
@@ -197,7 +206,9 @@ pub fn create_key_file(key_file: &str, hash_alg: &str) -> io::Result<FirmwareRaw
                         "unsupported Named Curve",
                     ));
                 }
-                if key.subject_public_key.as_bytes()[0] != 0x04 {
+                if key.subject_public_key.as_bytes()[0] != 0x04
```
jyao1, line 209: what is 0x04 ?

td-shim-tools/src/enroller.rs
gaojiaqi7: The extra byte is used to mark whether the key is compressed or not

td-shim-tools/src/enroller.rs
gaojiaqi7: 0x04 means the key is uncompressed

td-shim-tools/src/enroller.rs
jyao1: I dont see there is such definition in https://github.com/confidential-containers/td-shim/blob/main/doc/secure_boot.md.

td-shim-tools/src/enroller.rs
jyao1: I dont see there is such definition in https://github.com/confidential-containers/td-shim/blob/main/doc/secure_boot.md.

gaojiaqi7: It does not exist in the td shim binary, the key filled into `Signature Block` is 48 + 48 bytes.

jyao1: Please add detail description on the expectation of the key data.

td-shim-tools/src/enroller.rs
jyao1: This is very confusing. a P384 pub key size should be 0x96.
If you add other byte, then the meaning is not P384 pub key size.

td-shim-tools/src/enroller.rs
jodh-intel: Also, what sort of compression is being used here? Could that ever change? If so, a `const` naming the type and version of compression algorithm represented by `0x04` would be useful.

gaojiaqi7: The compressed form is shorter than the uncompressed one, for ECDSA P384, the compressed public key is encoded in 49 bytes and uncompressed public key is endcoded in 97 bytes (both include a one-byte prefix), the compression form leaves out half of the coordinate value which can be calculated according to the fixed curve and the prefix. According to the [Elliptic Curve Cryptography Subject Public Key Information](https://www.rfc-editor.org/rfc/rfc5480#section-2.2), the uncompressed form is indicated by 0x04 and the compressed form is indicated by either 0x02 or 0x03, and it will not change.

jyao1: I think to support uncompressed version only is OK.
My point is:
1) Please define 0x04 as a meaningful macro/const value so that people know what is is. A magic number is not acceptable.
2) Please define ECDSA-384 key as 96 byte. If the extra byte is 0x4 as uncompressed indicator, it should be parsed separately. Defining ECDSA-384 key as 97 bytes is very confusing.

jyao1 (approved):

Issue #380, opened 2022-08-18
[MIRAI] Possible attempt to add with overflow in hob.rs


liuw1:
warning: Inline assembly code cannot be analyzed by MIRAI.
   --> td-uefi-pi/src/hob.rs:678:14
    |
678 |     unsafe { core::arch::asm!("lfence") }
    |              ^^^^^^^^^^^^^^^^^^^^^^^^^^

warning: possible attempt to add with overflow
   --> td-uefi-pi/src/pi/hob.rs:155:13
    |
155 |             self.alloc_descriptor.memory_base_address + self.alloc_descriptor.memory_length - 1,
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

warning: possible attempt to add with overflow
   --> td-uefi-pi/src/pi/hob.rs:187:13
    |
187 |             self.physical_start + self.resource_length - 1,
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

warning: possible attempt to add with overflow
   --> td-uefi-pi/src/pi/hob.rs:270:13
    |
270 |             self.base_address + self.length - 1
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

liuw1:
I will take it

PR #387, opened 2022-08-23
td-uefi-pi: Replace to dump the length to avoid overflow

liuw1:
Fix: #380


jodh-intel (approved):
Thanks @liuw1.

jyao1 (approved):

Issue #383, opened 2022-08-18
[MIRAI] Possible attempt to subtract with overflow in page_table.rs


liuw1:
warning: possible attempt to subtract with overflow
   --> td-paging/src/page_table.rs:111:9
    |
111 |         sz -= mapped_size;
    |         ^^^^^^^^^^^^^^^^^

liuw1:
I will take it

PR #389, opened 2022-08-23
td-paging: Use checked_sub() to avoid overflow in page_table.rs

liuw1:
Fix: #383

### DIFF
--- a/td-paging/src/page_table.rs
+++ b/td-paging/src/page_table.rs
@@ -108,7 +108,7 @@ pub fn create_mapping_with_flags(
             Size4KiB::SIZE
         };
 
-        sz -= mapped_size;
+        sz = sz.checked_sub(mapped_size).ok_or(Error::InvalidArguments)?;
         pa += mapped_size;
         va += mapped_size;
     }


td-paging/src/page_table.rs
```diff
@@ -108,7 +108,7 @@ pub fn create_mapping_with_flags(
             Size4KiB::SIZE
         };
 
-        sz -= mapped_size;
+        sz = sz.checked_sub(mapped_size).unwrap();
```
jodh-intel, line 111: This function returns a `Result<()>` so couldn't this be:

```suggestion
        sz = sz.checked_sub(mapped_size)?;
```
```suggestion
        sz = sz.checked_sub(mapped_size).unwrap();
```

liuw1: 111 | |         sz = sz.checked_sub(mapped_size)?;
    | |                                         ^ cannot use the `?` operator in a function that returns `core::result::Result<(), Error>`

gaojiaqi7: hi @liuw1 , should we return error or unwrap here? Why not change `sz` as a signed integer. If we don't accept `sz` less than `0x1000`, we should check it at https://github.com/confidential-containers/td-shim/blob/main/td-paging/src/page_table.rs#L84

liuw1: Resolved

Issue #376, opened 2022-08-16
[MIRAI] Potential arithmetic overflow in td-shim/memory.rs


liuw1:
warning: possible attempt to add with overflow
   --> td-shim/src/bin/td-shim/memory.rs:284:21
    |
284 |                     new.physical_start += to_be_accepted;
    |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

warning: possible attempt to subtract with overflow
   --> td-shim/src/bin/td-shim/memory.rs:285:21
    |
285 |                     new.resource_length -= to_be_accepted;
    |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

warning: possible attempt to add with overflow
   --> td-shim/src/bin/td-shim/memory.rs:286:38
    |
286 |                     resources.insert(idx + 1, new);
    |                                      ^^^^^^^

warning: possible attempt to subtract with overflow
   --> td-shim/src/bin/td-shim/memory.rs:298:21
    |
298 |                     to_be_accepted -= size;
    |                     ^^^^^^^^^^^^^^^^^^^^^^

liuw1:
Please assign to jiaqi

gaojiaqi7:
> warning: possible attempt to add with overflow --> td-shim/src/bin/td-shim/memory.rs:284:21 | 284 | new.physical_start += to_be_accepted; | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>
This can be ensured by checking the validity of the resource hobs

> warning: possible attempt to subtract with overflow --> td-shim/src/bin/td-shim/memory.rs:285:21 | 285 | new.resource_length -= to_be_accepted; | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>
the `to_be_accepted` here is less than or equal to the `resource_length`

> warning: possible attempt to add with overflow --> td-shim/src/bin/td-shim/memory.rs:286:38 | 286 | resources.insert(idx + 1, new); | ^^^^^^^
> 
It is impossible for `idx` to overflow since we don't so much memory to hold the `usize::MAX * size_of::<ResourceDescription>()`

> warning: possible attempt to subtract with overflow --> td-shim/src/bin/td-shim/memory.rs:298:21 | 298 | to_be_accepted -= size; | ^^^^^^^^^^^^^^^^^^^^^^
>
[...]

PR #391, opened 2022-08-24
Check the validity of the physical memory resources

gaojiaqi7:


td-shim/src/bin/td-shim/memory.rs
```diff
@@ -286,6 +287,16 @@ impl<'a> Memory<'a> {
...
         use td_uefi_pi::pi;
 
+        // The physical address must not exceed the shared mask (the last bit of GPAW).
+        let max_phys_addr = resources
+            .iter()
+            .map(|resource| resource.physical_start + resource.resource_length - 1)
+            .max()
+            .unwrap();
+        if max_phys_addr > td::get_shared_page_mask() {
+            panic!("Invalid physical address for TDX");
```
jodh-intel, line 297: - Would it be useful to log `max_phys_addr` in the panic call?
- If so, would it be useful to log the resource index / indices of the invalid resources too?

jodh-intel (approved):
Thanks @gaojiaqi7.

gaojiaqi7:
These checks can avoid the overflow in https://github.com/confidential-containers/td-shim/issues/376

jyao1 (approved):

PR #395, opened 2022-08-31
td-shim: remove resource hobs for executable payload

gaojiaqi7:


gaojiaqi7:
E820 table is used to pass the memory information

jodh-intel (commented):
Thanks @gaojiaqi7. Is there any way you could add a unit test to assert that the hob _does not_ contain these resources (now)?

gaojiaqi7:
Hi @jodh-intel , we can't add unit test here since the `build_payload_hob` writes directly into raw memory address `memory.layout.runtime_acpi_base` which depends on runtime environment

jodh-intel:
@gaojiaqi7 - It should be possible to unit test the code by using some more config magic in `RuntimeMemoryLayout::new()` like this:

```rust
#[cfg(test)]
let runtime_acpi_base = fake_memory_area;

#[cfg(not(test))]
let runtime_acpi_base = real_memory_area;
```

But that might be better for a future PR maybe (?) as this one is only removing a few lines of code :smile:

gaojiaqi7:
Yes. Maybe we can also check the validity in the integration test, and I'd like to make it a new PR

jyao1 (approved):

PR #397, opened 2022-08-31
td-layout: correct the typo

gaojiaqi7:

### DIFF
--- a/td-layout/src/lib.rs
+++ b/td-layout/src/lib.rs
@@ -31,7 +31,7 @@ use runtime::*;
 // Minimal memory size to build the runtime layout.
 #[cfg(feature = "boot-kernel")]
 pub const MIN_MEMORY_SIZE: u64 = (TD_PAYLOAD_ACPI_SIZE
-    + TD_PAYLOAD_PARTIAL_ACCEPT_MEMORY_SIZE
+    + TD_PAYLOAD_UNACCEPTED_MEMORY_BITMAP_SIZE
     + TD_PAYLOAD_PAGE_TABLE_SIZE
     + TD_PAYLOAD_EVENT_LOG_SIZE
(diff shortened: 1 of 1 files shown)


td-layout/src/lib.rs
```diff
@@ -31,7 +31,7 @@ use runtime::*;
 // Minimal memory size to build the runtime layout.
 #[cfg(feature = "boot-kernel")]
 pub const MIN_MEMORY_SIZE: u64 = (TD_PAYLOAD_ACPI_SIZE
-    + TD_PAYLOAD_PARTIAL_ACCEPT_MEMORY_SIZE
+    + TD_PAYLOAD_UNACCEPTED_MEMORY_BITMAP_SIZE
```
jodh-intel, line 34: These `const`'s have radically different values:

```rust
pub const TD_PAYLOAD_PARTIAL_ACCEPT_MEMORY_SIZE: u32 = 0x10000000;
pub const TD_PAYLOAD_UNACCEPTED_MEMORY_BITMAP_SIZE: u32 = 0x40000;
```

... so although it might seem trivial, but how about a unit test to assert the expected value for `MIN_MEMORY_SIZE`? The fact that it was previously incorrect suggests it's easy to calculate it incorrectly so if we had a unit test, we'd always have that "safety net" to detect the change if we ever did need to change its value.

jodh-intel (commented):
Thanks @gaojiaqi7 - a couple of comments...

td-layout/src/lib.rs
gaojiaqi7: thanks @jodh-intel , have added more unit tests for the runtime layout.

jodh-intel (approved):
Thanks @gaojiaqi7.

lgtm

jyao1 (approved):

Issue #348, opened 2022-07-28
Disable log message for release build


gaojiaqi7:

gaojiaqi7:
https://github.com/confidential-containers/td-shim/pull/400

PR #400, opened 2022-08-31
Disable log message in release build

gaojiaqi7:
Log message is disabled in td-shim release version and reserved in the
td-payload and test-td-payload.

### DIFF
--- a/sh_script/integration_tdx.sh
+++ b/sh_script/integration_tdx.sh
@@ -187,18 +187,7 @@ launch_td_test_payload() {
 test_secure_boot() {
     echo "-- secure boot test"
     local time_out=10
-    if [[ ${firmware} == *normal* ]]
-    then
-        local key_str="Starting td-payload hob"
-    elif [[ ${firmware} == *mismatch-pubkey* ]]
-    then
-        local key_str="Verification fails: InvalidPublicKey"
-    elif [[ ${firmware} == *unsigned* ]]
-    then
-        local key_str="Secure Boot: Cannot read verify header from payload binary"
-    else
-        die "The firmware name ${firmware} is not suitable for secure boot test"
-    fi  
+    local key_str="Starting td-payload hob"
     
     nohup ${qemu_tdx_path} -accel kvm \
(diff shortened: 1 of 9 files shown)


gaojiaqi7:
https://github.com/confidential-containers/td-shim/issues/348

gaojiaqi7:
hi @jyao1 , current implementation reduces the size of release version image from `143 KB` to `81 KB`. After I deleted all calls to the `log`s, the size of the binary is reduced to `75K`, but this reduction should be due to the string literal, do you think it is neccessary for us to do more optimizations? I am not sure if the string literals can be optimized if we use another macro to replace the `log` with an empty implementation.

jodh-intel:
@gaojiaqi7, @jyao1 - you might want to consider tweaking the features in the manifest file. If you set [`release_max_level_off`](https://docs.rs/log/0.4.17/log/#compile-time-filters), all the logging code would be optimised out at compilation time for release builds, or if that's too brutal, `release_max_level_warn` might work?

gaojiaqi7:
thanks @jodh-intel , [release_max_level_off](https://docs.rs/log/0.4.17/log/#compile-time-filters) is good for us, we need to disable all the log message in release build and optimize the code size as much as possible

jyao1:
I think we need use `release_max_level_off`. Dont use runtime check.

td-shim/src/bin/td-shim/main.rs
```diff
@@ -95,7 +95,13 @@ pub extern "win64" fn _start(
     info: usize,
 ) -> ! {
     // The bootstrap code has setup the stack, but only the stack is available now...
-    let _ = td_logger::init();
+    let log_level = if cfg!(debug_assertions) {
```
jyao1, line 98: I think we need use `release_max_level_off`. Dont use runtime check.

gaojiaqi7: Thanks, I have updated the PR

jyao1 (approved):

Issue #405, opened 2022-09-06
Move event log out from td-shim and enhance its API


gaojiaqi7:

jyao1:
Please add more detail on why this is needed.
This description just tells what, but no clue on why.

gaojiaqi7:
It is moved out from `td-shim` crate to facilitate other programs to use it to replay or query the event log, including `EventlogWriter`, `EventlogReader` and some TCG definitions

gaojiaqi7:
Completed by https://github.com/confidential-containers/td-shim/pull/406

PR #406, opened 2022-09-06
td-shim: move `event_log` out to `tcg` crate

gaojiaqi7:
- Move generic TCG definitions out to the `tcg` crate.
 - Keep TD-Shim related event log definitions in the `td-shim/event_log`.
 - Move the `CCEL` ACPI table definition to `td-shim/acpi`.


tcg/src/log.rs
```diff
@@ -0,0 +1,377 @@
...
+    offset: usize,
+    last: usize,
+    extender: Box<dyn Fn(&[u8; SHA384_DIGEST_SIZE], u32)>,
+}
+
+impl<'a> CcEventLogWritter<'a> {
+    pub fn new(
+        area: &mut [u8],
+        extender: Box<dyn Fn(&[u8; SHA384_DIGEST_SIZE], u32)>,
+    ) -> Option<CcEventLogWritter> {
```
jodh-intel, line 27: It might be more conventional to return a `Result<>` here?

tcg/src/log.rs
```diff
@@ -0,0 +1,377 @@
...
+    }
+
+    fn calculate_digest_and_extend(
+        &self,
+        hash_data: &[u8],
+        mr_index: u32,
+    ) -> [u8; SHA384_DIGEST_SIZE] {
+        let hash_value = ring::digest::digest(&ring::digest::SHA384, hash_data);
+        let hash_value = hash_value.as_ref();
+        assert_eq!(hash_value.len(), SHA384_DIGEST_SIZE);
```
jodh-intel, line 76: Just querying this assert. Would it be better to protect it with `#[cfg(test)]`, or should this be a `panic!()`, or should the fucntion return a `Result<>`?

jodh-intel (commented):
Thanks @gaojiaqi7 - a few comments...

tcg/src/log.rs
gaojiaqi7: It should always be `SHA384_DIGEST_SIZE` when we use `SHA384` to calculate the digest, is it good to use assert here?

jyao1:
Based upon https://github.com/confidential-containers/td-shim/issues/405, I recommend using "cc_event_log" as crate name instead of "tcg".

jyao1 (approved):

Issue #339, opened 2022-07-18
Need to extend the kernel/kernel parameter into RTMR


gaojiaqi7:
To measure the kernel and its parameters, we need to know their size in memory

gaojiaqi7:
Done by https://github.com/confidential-containers/td-shim/pull/407

PR #407, opened 2022-09-06
Measure and extend kernel binary and parameter

gaojiaqi7:
Fix the https://github.com/confidential-containers/td-shim/issues/339

### DIFF
--- a/td-layout/src/memslice.rs
+++ b/td-layout/src/memslice.rs
@@ -7,8 +7,8 @@ use crate::build_time::{
     TD_SHIM_PAYLOAD_BASE, TD_SHIM_PAYLOAD_SIZE,
 };
 use crate::runtime::{
-    KERNEL_BASE, KERNEL_SIZE, TD_HOB_BASE, TD_HOB_SIZE, TD_PAYLOAD_ACPI_SIZE,
-    TD_PAYLOAD_EVENT_LOG_SIZE, TD_PAYLOAD_MAILBOX_SIZE, TD_PAYLOAD_SIZE,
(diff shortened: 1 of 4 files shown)


td-layout/src/memslice.rs
```diff
@@ -55,6 +57,10 @@ pub fn get_mem_slice<'a>(t: SliceType) -> &'a [u8] {
             SliceType::TdHob => {
                 core::slice::from_raw_parts(TD_HOB_BASE as *const u8, TD_HOB_SIZE as usize)
             }
+            SliceType::KernelParameter => core::slice::from_raw_parts(
+                KERNEL_PARAM_BASE as *const u8,
+                KERNEL_PARAM_SIZE as usize,
```
jodh-intel, line 62: Presumably if the params occupy < `KERNEL_PARAM_SIZE` bytes, the rest of the space is guaranteed to be zero'd? Is it worth adding a test for this since it seems like the length could vary depending on workload?

jodh-intel (commented):
Thanks @gaojiaqi7 - a comment/question...

td-layout/src/memslice.rs
gaojiaqi7: VMM put the params into memory and td shim doesn't  known the actual params occupy, so we can't determine the size of memory should be zeroed here.

jyao1 (approved):

jodh-intel (approved):
Thanks @gaojiaqi7.

lgtm

jyao1 (approved):

Issue #403, opened 2022-08-31
[MIRAI] Possible index out of bounds in td-shim/src/lib.rs


liuw1:
warning: possible index out of bounds
   --> td-shim/src/lib.rs:119:5
    |
119 |     buf[0] = (data & 0xFF) as u8;
    |     ^^^^^^

warning: possible index out of bounds
   --> td-shim/src/lib.rs:120:5
    |
120 |     buf[1] = ((data >> 8) & 0xFF) as u8;
    |     ^^^^^^

warning: possible index out of bounds
   --> td-shim/src/lib.rs:121:5
    |
121 |     buf[2] = ((data >> 16) & 0xFF) as u8;
    |     ^^^^^^

liuw1:
**Reproduce command:**
cd td-shim && MIRAI_FLAGS="--diag=library" cargo mirai

PR #409, opened 2022-09-07
Fix out of bounds in td-shim/src/lib.rs

liuw1:
Fix: #403

### DIFF
--- a/td-shim-tools/src/enroller.rs
+++ b/td-shim-tools/src/enroller.rs
@@ -82,7 +82,7 @@ impl FirmwareRawFile {
             self.data.push(0);
         }
 
-        // Update lengh field
+        // Update length field
         write_u24(
             self.data.len() as u32,
             &mut self.data[20..23].try_into().unwrap(),
--- a/td-shim-tools/src/enroller.rs
+++ b/td-shim-tools/src/enroller.rs
@@ -3,7 +3,7 @@
 //
 // SPDX-License-Identifier: BSD-2-Clause-Patent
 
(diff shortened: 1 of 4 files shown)


td-shim-tools/src/enroller.rs
```diff
@@ -82,7 +83,10 @@ impl FirmwareRawFile {
         }
 
         // Update lengh field
```
jodh-intel, line 85: Not part of this PR, but a typo: "length".

td-shim/src/lib.rs
```diff
@@ -114,8 +114,8 @@ pub struct PayloadInfo {
 }
 
 /// Write three bytes from an integer value into the buffer.
-pub fn write_u24(data: u32, buf: &mut [u8]) {
-    assert!(data < 0xffffff);
+pub fn write_u24(data: u32, buf: &mut [u8; 3]) {
+    assert!(data <= 0xffffff);
```
jodh-intel, line 118: Could you add a unit test for this change?

jodh-intel (commented):
Thanks @liuw1 - a few comments...

td-shim-tools/src/enroller.rs
liuw1: Fixed

td-shim/src/lib.rs
liuw1: Sure, added.

jodh-intel (approved):
Thanks @liuw1.

lgtm

Issue #399, opened 2022-08-31
[MIRAI] Possible range start index {} out of range for slice of length {} in td-loader


liuw1:
warning: possible range start index {} out of range for slice of length {}
   --> td-loader/src/elf64.rs:493:30
    |
493 |         let current_bytes = &self.entries[offset..];
    |                              ^^^^^^^^^^^^^^^^^^^^^^
    |
note: related location
   --> /home/liuwei/.rustup/toolchains/nightly-2022-08-08-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/slice/index.rs:18:9
    |
18  |         index.index(self)
    |         ^^^^^^^^^^^^^^^^^
note: related location
   --> /home/liuwei/.rustup/toolchains/nightly-2022-08-08-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/slice/index.rs:390:13
    |
390 |             slice_start_index_len_fail(self.start, slice.len());
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

warning: possible range start index {} out of range for slice of length {}
   --> td-loader/src/elf64.rs:563:30
    |
563 |         let current_bytes = &self.entries[offset..];
    |                              ^^^^^^^^^^^^^^^^^^^^^^
    |
note: related location
   --> /home/liuwei/.rustup/toolchains/nightly-2022-08-08-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/slice/index.rs:18:9
    |
18  |         index.index(self)
    |         ^^^^^^^^^^^^^^^^^
note: related location
   --> /home/liuwei/.rustup/toolchains/nightly-2022-08-08-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/slice/index.rs:390:13
    |
390 |             slice_start_index_len_fail(self.start, slice.len());
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

warning: possible range start index {} out of range for slice of length {}
   --> td-loader/src/elf64.rs:612:30
    |
612 |         let current_bytes = &self.entries[offset..];
    |                              ^^^^^^^^^^^^^^^^^^^^^^
    |
note: related location
[...]

Issue #398, opened 2022-08-31
[MIRAI] Potential arithmetic overflow in td-loader


liuw1:
warning: possible attempt to multiply with overflow
   --> td-loader/src/elf.rs:68:18
    |
68  |     for reloc in elf.relocations()? {
    |                  ^^^^^^^^^^^^^^^^^
    |
note: related location
   --> td-loader/src/elf64.rs:782:28
    |
782 |         let dynamic_info = self.dynamic_info()?;
    |                            ^^^^^^^^^^^^^^^^^^^
note: related location
   --> td-loader/src/elf64.rs:741:23
    |
741 |         for header in self.program_headers() {
    |                       ^^^^^^^^^^^^^^^^^^^^^^
note: related location
   --> td-loader/src/elf64.rs:491:22
    |
491 |         let offset = self.index * ENTRY_SIZE;
    |                      ^^^^^^^^^^^^^^^^^^^^^^^

warning: possible attempt to subtract with overflow
  --> td-loader/src/elf.rs:83:9
   |
83 |         (top - bottom) as u64,
   |         ^^^^^^^^^^^^^^

warning: possible attempt to multiply with overflow
   --> td-loader/src/elf.rs:91:15
    |
91  |     for sh in elf.section_headers() {
    |               ^^^^^^^^^^^^^^^^^^^^^
    |
note: related location
   --> td-loader/src/elf64.rs:561:22
    |
561 |         let offset = self.index * ENTRY_SIZE;
    |                      ^^^^^^^^^^^^^^^^^^^^^^^

warning: possible attempt to multiply with overflow
   --> td-loader/src/elf.rs:104:15
    |
104 |     for sh in elf.section_headers() {
    |               ^^^^^^^^^^^^^^^^^^^^^
    |
note: related location
   --> td-loader/src/elf64.rs:561:22
    |
561 |         let offset = self.index * ENTRY_SIZE;
    |                      ^^^^^^^^^^^^^^^^^^^^^^^

warning: possible attempt to multiply with overflow
   --> td-loader/src/elf.rs:117:15
    |
117 |     for sh in elf.section_headers() {
    |               ^^^^^^^^^^^^^^^^^^^^^
    |
note: related location
   --> td-loader/src/elf64.rs:561:22
    |
561 |         let offset = self.index * ENTRY_SIZE;
    |                      ^^^^^^^^^^^^^^^^^^^^^^^

warning: possible attempt to multiply with overflow
[...]

PR #410, opened 2022-09-07
td-loader: Fix potential arithmetic overflow and index out of range warnings

liuw1:
Fix: #398
Fix: #399


jodh-intel (approved):
Thanks @liuw1. It would be geat if you could add some unit tests for this change.

liuw1:
> Thanks @liuw1. It would be geat if you could add some unit tests for this change.

Good suggestion, I will plan to add some unit test later. Thanks!

jyao1 (approved):

jyao1:
> Good suggestion, I will plan to add some unit test later. Thanks!

@liuw1 Please submit an issue to track it.

Issue #404, opened 2022-08-31
[MIRAI] Potential arithmetic overflow in td-shim


liuw1:
warning: possible attempt to subtract with overflow
   --> td-shim/src/acpi.rs:108:13
    |
108 |             (self.header.length as usize - size_of::<GenericSdtHeader>()) / size_of::<u64>();
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

warning: possible attempt to add with overflow
   --> td-shim/src/event_log.rs:351:17
    |
351 |                 offset += CC_EVENT_HEADER_SIZE;
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

liuw1:
**Reproduce command:**
cd td-shim && MIRAI_FLAGS="--diag=library" cargo mirai

liuw1:
Both are false positive:
--> td-shim/src/acpi.rs:108:13
self.header.length is always >= size_of::<GenericSdtHeader>(). Xsdt init with length size_of::<GenericSdtHeader>().

--> td-shim/src/event_log.rs:351:17
If the code goes to this condition means read cc_event_header successfully, so it cannot be overflow.

PR #411, opened 2022-09-07
td-shim: Fix potantial arithmetic overflow warning

liuw1:
Fix: #404

### DIFF
--- a/td-shim/src/acpi.rs
+++ b/td-shim/src/acpi.rs
@@ -194,4 +194,29 @@ mod tests {
         xsdt.checksum();
         assert_eq!(xsdt.header.checksum, CHECK_SUM);
     }
+
+    #[test]
+    fn test_xsdt_add_table() {
+        let mut xsdt = Xsdt::new();
+
+        xsdt.add_table(100);
+        let first_table = xsdt.tables[0];
+        assert_eq!(first_table, 100);
+
+        // length < size_of::<GenericSdtHeader>(), add table fail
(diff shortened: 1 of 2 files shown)


td-shim/src/acpi.rs
```diff
@@ -104,13 +104,19 @@ impl Xsdt {
...
+            );
         } else {
-            log::error!("too many ACPI tables, max {}", ACPI_TABLES_MAX_NUM);
+            let table_num =
+                (self.header.length as usize - size_of::<GenericSdtHeader>()) / size_of::<u64>();
+            if table_num < ACPI_TABLES_MAX_NUM {
+                self.tables[table_num] = addr as u64;
+                self.header.length += size_of::<u64>() as u32;
+            } else {
+                log::error!("too many ACPI tables, max {}", ACPI_TABLES_MAX_NUM);
```
jodh-intel, line 118: This is hard to unit test so I do wonder if either `add_table()` should return a `Result<>`, or if we need something like:

```suggestion
                #[cfg(not(test))]
                log::error!("too many ACPI tables, max {}", ACPI_TABLES_MAX_NUM);

                #[cfg(test)]
                panic!("too many ACPI tables, max {}", ACPI_TABLES_MAX_NUM);
```

You could then use a `#[should_panic(...)]` on a unit test to assert the expected behaviour.

liuw1: I added test case to check the table value with different add_table() scenarios.

jodh-intel (approved):
Thanks @liuw1.

lgtm

jyao1 (approved):

Issue #396, opened 2022-08-31
[MIRAI] Potential arithmetic overflow in td-layout


liuw1:
warning: possible attempt to subtract with overflow
  --> td-layout/src/lib.rs:99:13
   |
99 |             current_base -= TD_PAYLOAD_UNACCEPTED_MEMORY_BITMAP_SIZE as u64;
   |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

warning: possible attempt to add with overflow
   --> td-layout/src/metadata.rs:228:21
    |
228 |                     bfv_cnt += 1;
    |                     ^^^^^^^^^^^^

warning: possible attempt to add with overflow
   --> td-layout/src/metadata.rs:268:21
    |
268 |                     hob_cnt += 1;
    |                     ^^^^^^^^^^^^

warning: possible attempt to add with overflow
   --> td-layout/src/metadata.rs:312:21
    |
312 |                     perm_mem_cnt += 1;
    |                     ^^^^^^^^^^^^^^^^^

warning: possible attempt to add with overflow
   --> td-layout/src/metadata.rs:333:21
    |
333 |                     payload_cnt += 1;
    |                     ^^^^^^^^^^^^^^^^

warning: possible attempt to add with overflow
   --> td-layout/src/metadata.rs:353:21
    |
353 |                     payload_param_cnt += 1;
    |                     ^^^

gaojiaqi7:
> --> td-layout/src/lib.rs:99:13
> |
> 99 | current_base -= TD_PAYLOAD_UNACCEPTED_MEMORY_BITMAP_SIZE as u64;
> | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

hi, can you help to try with the https://github.com/confidential-containers/td-shim/pull/397 to see if this still a warning?

PR #412, opened 2022-09-07
td-layout: Fix potential arithmetic overflow

liuw1:
Fix: #396

### DIFF
--- a/td-layout/src/metadata.rs
+++ b/td-layout/src/metadata.rs
@@ -225,6 +225,9 @@ impl TdxMetadata {
                 TDX_METADATA_SECTION_TYPE_BFV => {
                     // A TD-Shim shall include at least one BFV and the reset vector shall be inside
                     // of BFV. The RawDataSize of BFV must be non-zero.
+                    if bfv_cnt == i32::MAX {
+                        return false;
+                    }
                     bfv_cnt += 1;
                     if section.raw_data_size == 0 {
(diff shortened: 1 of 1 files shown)


jodh-intel (commented):
Hi @liuw1 - Could you add some unit tests for this change?

liuw1:
> Hi @liuw1 - Could you add some unit tests for this change?

Hi @jodh-intel It's hard to add unit test case to cover the scenario. But the MIRAI will report the warnings.

jodh-intel:
@liuw1 - Agreed. To make it easier (more reliable) to unit-test, you could create something like:

```rust
fn validate_sections(sections: &[TdxMetadataSection]) -> Result<()> {
    // ...

    Ok(())
}
```

And then rework `is_valid_sections()` to return `true` if `validate_sections()` return `()`, or `false` on `Err()`. Then, you could reliably unit test `validate_sections()` as you could check the specified error returned.

jodh-intel (approved):
The change looks reasonable, but I'd still prefer some sort of UT to verify it.

jyao1 (approved):

jyao1:
@liuw1 , I approved it since it follows existing pattern to fix the issue.

Please consider using "fn validate_sections(sections: &[TdxMetadataSection]) -> Result<()> {" to check instead of bool.
Please file an issue for that.

PR #419, opened 2022-09-21
td-layout: Refactor funtion is_valid_sections()

liuw1:
Fix: #417

### DIFF
--- a/td-layout/src/metadata.rs
+++ b/td-layout/src/metadata.rs
@@ -158,6 +158,11 @@ impl Default for TdxMetadataGuid {
     }
 }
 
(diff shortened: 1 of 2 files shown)


td-layout/src/metadata.rs
```diff
@@ -158,6 +158,11 @@ impl Default for TdxMetadataGuid {
     }
 }
 
+#[derive(Debug)]
+pub enum InvalidSectionError {
```
jyao1, line 162: should it be TdxMetadataError?

liuw1: Changed.

PR #422, opened 2022-10-08
Fix CI build issues

liuw1:
Correct the traget for PE format payload build


jyao1:
@gaojiaqi7 , could you please take a look? We should fix all -uefi to -none. Why CI failed?

jyao1 (approved):

PR #425, opened 2022-10-11
td-loader: get .init_array with the PT_DYNAMIC segment

gaojiaqi7:
Section header table is not loaded into memory, runtime initialization
program can use the `PT_DYNAMIC` segment to find the `.init_array` or
`.finit_array`.

Update the implementation of the `parse_init_array_section` function
and load all the program segment into memory.


gaojiaqi7:
Fix https://github.com/confidential-containers/td-shim/issues/427

jyao1 (approved):

Issue #294, opened 2022-05-31
use sha2 and/or openssl/mbedtls to replace ring


jyao1:
Current td-shim need use SHA384 to calculate the MR extend.

Since ring upstream is blocked, we should use sha2 to replace ring.

Rust crypto provides https://github.com/RustCrypto/hashes. SHA384 is available in https://crates.io/crates/sha2

gaojiaqi7:
We still use ring to verify the payload signature, is there a substitute for this?

jyao1:
That is secure boot specific, right? Let's first put ring to secure-boot feature.

jyao1:
For secure boot, we may consider https://crates.io/crates/openssl

jodh-intel:
Related: https://github.com/confidential-containers/community/issues/44.

ariel-adam:
@jyao1 is this issue still relevant or can be closed? 
If it's still relevant to what release do you think we should map it to (mid-November, end-December, mid-February etc...)?

dcmiddle:
@jyao1 can you clarify:
> Since ring upstream is blocked, we should use sha2 to replace ring.

Is there a blocked issue or something we can track?

dcmiddle:
@jyao1 is this issue resolved now via #428?

PR #428, opened 2022-10-13
cc-measurement: replace `ring` with `sha2`

gaojiaqi7:
For https://github.com/confidential-containers/td-shim/issues/294, use crate sha2 to calculate the hash digest.


jyao1:
Can we make it a feature?

jyao1 (approved):

Issue #446, opened 2022-11-07
Need to fix release.yml


liuw1:
https://github.com/confidential-containers/td-shim/actions/runs/3408280998

PR #447, opened 2022-11-07
github: Fix release workflow

fidencio:
While the release workflow looks for the 'final-pe-boot-kernel.bin' binary, the binary name should actually be 'final-boot-kernel.bin'

Fixes: #446

Signed-off-by: Fabiano Fidêncio <fabiano.fidencio@intel.com>

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -73,15 +73,15 @@ jobs:
           asset_name: final-pe.bin
           asset_content_type: application/octet-stream
 
-      - name: Upload final-pe-boot-kernel.bin
-        id: upload_release_final_pe_boot_kernel_bin
+      - name: Upload final-boot-kernel.bin
+        id: upload_release_final_boot_kernel_bin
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -73,15 +73,15 @@ jobs:
           asset_name: final-pe.bin
           asset_content_type: application/octet-stream
 
-      - name: Upload final-pe-boot-kernel.bin
+      - name: Upload final-boot-kernel.bin
         id: upload_release_final_pe_boot_kernel_bin
```
liuw1, line 77: The id also need to be corrected. 😊

liuw1 (changes requested):

.github/workflows/release.yaml
fidencio: Updated, nice catch!

liuw1 (approved):

jyao1 (approved):

Issue #449, opened 2022-11-09
Need support `PermMem` tdvf section type


gaojiaqi7:
If PermMem section exists, TD will own the memory allocation and TD HOB should be ignored.

gaojiaqi7:
Completed by https://github.com/confidential-containers/td-shim/pull/450

PR #450, opened 2022-11-11
Support `PermMem` metadata section and user-defined metadata

gaojiaqi7:
Align the implementation with https://github.com/confidential-containers/td-shim/blob/main/doc/tdshim_spec.md#metadata-definition


gaojiaqi7:
Fix https://github.com/confidential-containers/td-shim/issues/449

td-shim/src/bin/td-shim/info.rs
```diff
@@ -0,0 +1,197 @@
...
+impl<'a> ShimInfo<'a> {
+    // Validate the metadata and get the basic infomation from
+    // it if any
+    pub fn new() -> Option<Self> {
+        let metadata_offset = unsafe { *((u32::MAX - TDX_METADATA_OFFSET + 1) as *const u32) };
+        if metadata_offset >= TD_SHIM_FIRMWARE_SIZE
+            || metadata_offset < size_of::<Guid>() as u32
+            || metadata_offset > TD_SHIM_FIRMWARE_SIZE - size_of::<TdxMetadataDescriptor>() as u32
+        {
+            error!("Invalid TDX Metadata offset\n");
```
jyao1, line 43: I remember we support both OFFSET style and GUID style. Should we support both cases here?

td-shim/src/bin/td-shim/info.rs
```diff
@@ -0,0 +1,197 @@
+// Copyright (c) 2022 Intel Corporation
+//
+// SPDX-License-Identifier: BSD-2-Clause-Patent
+
+use alloc::vec::Vec;
```
jyao1, line 5: Can we rename file to be shim_info?

td-shim/src/bin/td-shim/info.rs
```diff
@@ -0,0 +1,197 @@
...
+    // If metadata contains the `TD_HOB` section, TD-Shim
+    // can get additional information from TD HOB
+    pub td_hob: Option<&'static [u8]>,
+
+    // If metadata contains one/more `PermMem` sections,
+    // TD-Shim should ignore the memory information in TD HOB.
+    metadata_has_perm: bool,
+
+    pub memory: Vec<ResourceDescription>,
+    pub acpi_tables: Vec<&'a [u8]>,
```
jyao1, line 29: I feel it is weird to put acpi_tables here.

I assume ShimInfo means the metadata / memory related static information, right?

Or you want to create a structure to hold Runtime information ?

td-shim/src/bin/td-shim/info.rs
gaojiaqi7: There is always an `OFFSET` at `4GiB - 0x20` no matter if there is a GUID chain or not

td-shim/src/bin/td-shim/info.rs
gaojiaqi7: Sure

td-shim/src/bin/td-shim/info.rs
gaojiaqi7: I agree that we should separate the different kind of informations. The `ShimInfo` is just derived from original `TdHobInfo` structure and combined `1` and `2`

jyao1 (approved):

PR #451, opened 2022-11-14
td-shim: disable crate `log` in td-shim library

gaojiaqi7:
Any other crates that depend on `td-shim` will derive the feature of the `log` crate, so make it optional and enable it with `main`.

### DIFF
--- a/td-shim/src/acpi.rs
+++ b/td-shim/src/acpi.rs
@@ -105,17 +105,13 @@ impl Xsdt {
 
     pub fn add_table(&mut self, addr: u64) {
         if self.header.length < size_of::<GenericSdtHeader>() as u32 {
-            log::error!(
-                "Invalid header: Xsdt header length should not be less than generic header size"
-            );
+            return;
         } else {
             let table_num =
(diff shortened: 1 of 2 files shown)


gaojiaqi7:
Will handle the error in `acpi.rs` in another pull request

gaojiaqi7:
Fix: https://github.com/confidential-containers/td-shim/issues/452

td-shim/src/acpi.rs
```diff
@@ -105,17 +105,13 @@ impl Xsdt {
 
     pub fn add_table(&mut self, addr: u64) {
         if self.header.length < size_of::<GenericSdtHeader>() as u32 {
-            log::error!(
-                "Invalid header: Xsdt header length should not be less than generic header size"
-            );
+            return;
```
jyao1, line 108: If this is expected, then we need return error.

If this is unexpected, then we need use assert.

jyao1: if you have another issue to resolve, please add issue number here.

jyao1 (approved):

PR #464, opened 2022-12-07
td-shim-tools: Introduce a new tool to generate runtime layout.

xiaoyuxlu:
This tool is for td-payload and td-shim to generate runtime layout config `td-layout/src/runtime.rs`.

1. Generate runtime.rs from `devtools/td-layout-config/config.json` file.  
    The generated output is compatible with `td-layout-config`
    ```
    ./target/debug/td-shim-layout-builder -p devtools/td-layout-config/config.json
    ```
2. Generate runtime.rs file from jsonfile. 
    ```
    ./target/debug/td-shim-layout-builder -p -t json td-shim-tools/src/bin/td-shim-layout-builder/runtime.json
    ```


td-shim-tools/src/bin/td-shim-layout-builder/main.rs
```diff
@@ -0,0 +1,62 @@
...
+
+use clap::{Parser, ValueEnum};
+use td_shim_tools::layout_builder::{
+    self, parse_csv, parse_json, parse_layout_config, MemoryRegions,
+};
+
+#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, ValueEnum)]
+pub enum ConfigType {
+    /// csv format config
+    CSV,
```
jyao1, line 13: JSON is enough. Why we need CSV?

xiaoyuxlu: CSV format is optional.  Json is also supported.
Memory layout is more like to a table. CSV is designed for tables.
To a same layout CSV format 14 lines are needed, but Json format needs 54 lines. It's more clearer.

jyao1: Why we have to support both formats?

I think JSON is enough. Please remove CSV format at all.

xiaoyuxlu: done

jyao1 (approved):

PR #468, opened 2022-12-11
improve td-payload crate

gaojiaqi7:


td-payload/src/arch/x86_64/pci.rs
```diff
@@ -0,0 +1,486 @@
+// Copyright (c) 2022 Intel Corporation
+//
+// SPDX-License-Identifier: BSD-2-Clause-Patent
+
+use bitflags::bitflags;
+use core::convert::From;
+
+pub const PCI_CONFIGURATION_ADDRESS_PORT: u16 = 0xCF8;
+pub const PCI_CONFIGURATION_DATA_PORT: u16 = 0xCFC;
+pub const PCI_EX_BAR_BASE_ADDRESS: u64 = 0xE0000000u64;
```
jyao1, line 10: why we have such code?

td-exception/src/interrupt.rs
```diff
@@ -424,7 +431,40 @@ interrupt_no_error!(virtualization, stack, {
...
+        );
+
+        // SSP -> Return address of current function | SSP | LIP | CS
+        let lip_ptr = ssp + 0x10;
+        let lip = stack.iret.rip;
+
+        // Enables the WRSSD/WRSSQ instructions by setting the `WR_SHSTK_E`
+        // to 1, then we can write the shadow stack
+        let mut msr_cet = Msr::new(MSR_IA32_S_CET);
+        msr_cet.write(msr_cet.read() | WR_SHSTK_E);
```
jyao1, line 456: We should not enable WR.

Please refer to https://github.com/tianocore/edk2/blob/master/MdePkg/Library/BaseLib/X64/LongJump.nasm#L49

gaojiaqi7: We need to change the `LIP` saved in `shadow stack memory`, not the shadow stack pointer. Change the shadow stack pointer doesn't affect the behavior returning from an exception (`iret`)

jyao1:
please dont introduce new allocator.

jyao1 (approved):

Issue #418, opened 2022-09-15
TD-Shim must warn payload how IDT is set up


dimakuv:
I am developing my ELF payload that TD-Shim loads.

## Security/usability issue

My ELF payload performs some `rdmsr`, `wrmsr`, `cpuid` instructions. Under Intel TDX, **only a predefined subset** of MSRs and CPUID leaves can be accessed natively with these instructions; the rest MSRs and CPUID leaves generate a #VE exception, and may be accessed by the #VE handler via `tdvmcall_rdmsr`, `tdvmcall_wrmsr` and `tdvmcall_cpuid`.

In the following, I'll take an example of CPUID. But the same problem relates to MSRs.

TD-Shim installs a #VE exception handler (more specifically, an ISR in the IDT table) during its initialization:  https://github.com/confidential-containers/td-shim/blob/d4cfe2ce56d2052219d3cd7efe878552424c4e21/td-exception/src/idt.rs#L106

This #VE exception handler simply forwards the request from the payload to the host VMM (via `tdvmcall_cpuid`): https://github.com/confidential-containers/td-shim/blob/d4cfe2ce56d2052219d3cd7efe878552424c4e21/td-exception/src/interrupt.rs#L404-L411

There is **no sanitization/verification** of the CPUID leaves. There is also no debug message or anything to inform the developer that a #VE exception was resolved, and the CPUID outputs should be considered **untrusted**.

The TD-Shim specification only says this:
> The expectation for OS is that: OS shall setup its own GDT/IDT/CR3 in the 4KB OS wakeup page immediately and jump to anywhere supported by OS page table.

First, this information is burried under the `Multiple Processor Support` section. Second, this information only refers to the OS kernel, but not to an "arbitrary payload".

I was confused and puzzled why seemingly "not allowed by Intel TDX" accesses to MSRs and CPUID leaves were working... It took me quite some time before I realized that TD-Shim **silently set it up for its own purposes**.

### Proposed solutions

1. Make a clear statement about the #VE exception handling (that it forwards MSR, CPUID, other requests to the host VMM) in the TD-Shim specification. Explain that the results are **untrusted and must be sanitized**.
2. Add a debug message to the #VE exception handler. Explain that the results are **untrusted and must be sanitized**.
3. Somehow force the payload to set up its own GDT/IDT. Well, maybe just spell it out in the TD-Shim specification.

jyao1:
Good finding.
1. Agree. We can add more comment there.
[...]

PR #473, opened 2022-12-21
Reset IDT before transfer to payload

gaojiaqi7:
Fix https://github.com/confidential-containers/td-shim/issues/418

 - Set an empty IDT before jump to payload
   Two pages are reserved for exception handling before payload sets its own IDT. Descriptor table is in the first page and the exception hanlder is put in the second page.

    An empty exception handler is set which simply performs a `hlt` via TDVMCALL.

 - Set IDT for application processors
   Move mailbox relocation after setting IDT for APs.


doc/tdshim_spec.md
```diff
@@ -441,6 +441,9 @@ second parameter.
    loaded.
  * All rest are same as above.
 
+Before TD Shim hands over control to payload, TD Shim will reset the Interrupt
```
jyao1, line 444: I dont think TD shim should reset IDT. It will still have an IDT point to reserved memory.

The payload should set its own IDT, same as other GDT, page table, etc.

td-shim/src/bin/td-shim/asm/exception.asm
```diff
@@ -0,0 +1,32 @@
...
+    mov         rcx, TDVMCALL_EXPOSE_REGS_MASK
+    mov         r10, 0
+    mov         r11, INSTRUCTION_HLT
+    mov         r12, 0
+    mov         r13, 0
+    mov         r14, 0
+    mov         r15, 0
+    # TDVMCALL
+    .byte       0x66, 0x0f, 0x01, 0xcc
+    ud2
```
jyao1, line 28: please use `jmp $`

gaojiaqi7: `jmp $` can't be used in global assembly, put a symbol before the `jmp` instruction:
```
.exception_loop:
                jmp .exception_loop
```

jyao1: OK

jyao1 (approved):

Issue #461, opened 2022-12-06
need split the common memlayout to 2 different configuration


jyao1:
One is for boot-kernel (such as bzimage, vmlinuz), the other is for boot executable payload.

gaojiaqi7:
Shall we use two different image for linux and executable payload? or decide which one to use at runtime according to `PayloadInfoHob`

Issue #466, opened 2022-12-08
Remove partial_accept_memory_size in devtools/td-layout-config/config.json


xiaoyuxlu:
https://github.com/confidential-containers/td-shim/blob/07d0562531c00bc26faa794f68e0381c4d8a0d50/devtools/td-layout-config/config.json#L75

partial_accept_memory_size is for lazy accept.
td-layout-config/config.json is for generating td-layout files. 

They are independent. they should not be put together.

Issue #460, opened 2022-12-06
Remove `kernel` keyword


gaojiaqi7:

jyao1:
we can still use boot-kernel feature, but not use kernel keyword in mem layout.

jyao1:
please also remove `KERNEL`

PR #479, opened 2022-12-27
refactor `devtools/td-layout-config` and `td-layout` crates

gaojiaqi7:
### Issue related to this PR

 - Fix: https://github.com/confidential-containers/td-shim/issues/461

### Refactor the `devtools/td-layout-config`

1. Merge `td-shim-layout-builder` under `td-shim-tools` into `devtools/td-layout-config`

Make some optimizations to support memory type which can help to determine the location of memory regions (high-end or low-end of low memory) and to generate E820 table at runtime with the output rust format configurations.

The memory layout configuration file will look like this:

```
{
    "memory_regions": [
        {
            "name": "Bootloader",
            "size": "0x800000",
            "type": "Memory"
        },
        {
            "name": "TdHob",
            "size": "0x20000",
            "type": "Memory"
        },
        {
            "name": "PayloadParam",
            "size": "0x40000",
            "type": "Memory"
        }
        ...
        
        {
            "name": "EventLog",
            "size": "0x100000",
            "type": "Nvs"
        },
        ...
    ]
}
```

Add the image layout configuration to the new tool which is not supported by `td-shim-layout-builder`.

2. Split the image configuration and memory configuration

Linux payload and excutable payload will use one same image configuration but different memory layout, so we can split image and memory configuration for now.

### Update the `td-layout` used by runtime of `td-shim`

1. Use one `RuntimeLayout` defination for different layouts

Previous implementation of `RuntimeLayout` defines fixed members which can only be used for one layout. In this PR, an array is used to hold memory regions with address, size and type informations and it can be initialized with layout configuration at runtime.

2. Move `get_dynamic_mem_slice()` to `Memory`

`Memory` struct holds the information of memory regions and it will be more natural and convenient to get the slice of these regions from `Memory` struct.


gaojiaqi7:
#466 #461 #460

td-shim/src/bin/td-shim/memory.rs
```diff
@@ -399,7 +399,7 @@ pub fn cpu_get_memory_space_size() -> u8 {
     size_of_mem_space
 }
 
-#[cfg(test)]
+// #[cfg(test)]
```
jyao1, line 402: why comment it?

jyao1:
please split the patch to smaller ones. It is hard to review a refactoring patch with 66 files updated.

gaojiaqi7:
Split this PR into PR https://github.com/confidential-containers/td-shim/pull/519, PR https://github.com/confidential-containers/td-shim/pull/516, PR https://github.com/confidential-containers/td-shim/pull/517 and PR PR https://github.com/confidential-containers/td-shim/pull/518

Issue #465, opened 2022-12-08
Hardcode implementation in td-shim\src\bin\td-shim\linux\boot.rs need to be resolved


liuw1:
Both setup_header() and boot_kernel() have hardcode implementation.

PR #480, opened 2022-12-28
td-shim: add const definitions for Linux setup header

gaojiaqi7:
Fix: https://github.com/confidential-containers/td-shim/issues/465


td-shim/src/bin/td-shim/linux/kernel_param.rs
```diff
@@ -7,6 +7,12 @@ use zerocopy::{AsBytes, FromBytes};
 
 use td_shim::e820::E820Entry;
 
+pub const HDR_BOOT_FLAG: u16 = 0xaa55;
```
jyao1, line 10: please add a reference for the value.

jyao1 (approved):

Issue #462, opened 2022-12-06
remove stack in boot executable payload configure


jyao1:
The td-shim can switch stack and use the temp stack, and the payload should switch stack to its own region again.

gaojiaqi7:
Close with https://github.com/confidential-containers/td-shim/pull/481

PR #481, opened 2022-12-28
td-shim: do not switch new stack for executable payloads

gaojiaqi7:
- Remove the stack from `td-layout`
 - Don't reserve stack memory before transfer to payload
 - Replace `swich_stack_call` with normal function call


gaojiaqi7:
Fix https://github.com/confidential-containers/td-shim/issues/462

jyao1 (approved):

Issue #306, opened 2022-06-12
consider to enable IBT support


jyao1:
The rightly version can support [cf-protection](https://doc.rust-lang.org/nightly/unstable-book/compiler-flags/cf-protection.html#cf-protection).

`cf-protection=branch` maybe used to enable IBT.

gaojiaqi7:
Using `match` with lots of branches will trigger the control flow exception. It seems a rust compiler issue.
The status of rust cet support is still WIP..

PR #482, opened 2022-12-28
Add CET IBT support in library crates

gaojiaqi7:
Need to add `endbr64` in front of the position where the global asm functions starts.

- Replace the global asm functions in `td-exception` with the inline functions provided by crate `x86_64`
- Add `endbr64` in the global asm funcitons in `tdx-tdcall`


gaojiaqi7:
IBT support is tracked by the issue https://github.com/confidential-containers/td-shim/issues/306

tdx-tdcall/src/asm/tdcall.asm
```diff
@@ -19,6 +19,7 @@
 # )
 .global asm_td_call
 asm_td_call:
+        endbr64
```
jyao1, line 22: why we need endbr here?

gaojiaqi7: It seems global asm functions are called with an indirect CALL/JMP, so the program will crash if we do not add `endbr64` here.

jyao1: Would you please add some reference on that? I am a little surprised on why.

gaojiaqi7: This is the disassembly of https://github.com/confidential-containers/td-shim/blob/main/tdx-tdcall/src/lib.rs#L101-L103

```
0000000000000000 <td_call>:
   0:	f3 0f 1e fa          	endbr64 
   4:	48 83 ec 28          	sub    $0x28,%rsp
   8:	48 89 f9             	mov    %rdi,%rcx
   b:	ff 15 00 00 00 00    	call   *0x0(%rip)        # 11 <td_call+0x11>
  11:	48 83 c4 28          	add    $0x28,%rsp
  15:	c3                   	ret
```

jyao1: OK. It is indirect call.

jyao1 (approved):

jyao1:
please resolve conflict.

PR #501, opened 2023-01-17
td-payload: handle `retry` result of mapping GPA

gaojiaqi7:
Fix: https://github.com/confidential-containers/td-shim/issues/498

### DIFF
--- a/tdx-tdcall/src/lib.rs
+++ b/tdx-tdcall/src/lib.rs
@@ -62,6 +62,7 @@ pub const TDCALL_STATUS_PAGE_SIZE_MISMATCH: u64 = 0xC0000B0B00000001;
 
 // TDVMCALL completion status code
 const TDVMCALL_STATUS_SUCCESS: u64 = 0;
+const TDVMCALL_STATUS_RETRY: u64 = 1;
 
 // A public wrapper for use of asm_td_vmcall, this function takes a mutable reference of a
 // TdcallArgs structure to ensure the input is valid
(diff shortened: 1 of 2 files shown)


td-payload/src/arch/x86_64/dma.rs
```diff
@@ -3,16 +3,28 @@
 // SPDX-License-Identifier: BSD-2-Clause-Patent
 
 use crate::mm::SIZE_4K;
-use tdx_tdcall::tdx;
+use tdx_tdcall::{tdx, TdVmcallError};
 
 use super::paging::{clear_shared_bit, set_shared_bit};
 
+const MAP_GPA_MAX_RETRIES: u8 = 3;
```
jyao1, line 10: why retry 3 times?

gaojiaqi7:
Screenshot 2023-03-20 212728

gaojiaqi7:
Answer from KVM team:

> 1. Under what circumstances VMM may return this result?

When the specified range is too large. VMM can't complete the request in practical time. e.g. hundreds of GB.

It's up to VMM implementation.


> 2. The spec says guest TD must retry this operation when `TDG.VP.VMCALL_RETRY` is returned. Should guest TD set any maximum retries or timeout? or it should try again unconditionally, no matter how many times it has already tried?

Retry unconditionally. GHCI spec doesn't have any guide on the retry because it's attack surface for (malicious) VMM against guest TD.

If you want guest to implement guard, it's totally up to you how to protect it. probably it's heuristic protection.

mxu9:
It looks good to me.

jyao1 (approved):

PR #507, opened 2023-01-31
td-payload: Add feature coverage to enlarge DEFAULT_DMA_SIZE

liuw1:


td-payload/src/mm/layout.rs
```diff
@@ -5,7 +5,10 @@
 pub const DEFAULT_HEAP_SIZE: usize = 0x1000000;
 pub const DEFAULT_STACK_SIZE: usize = 0x800000;
 pub const DEFAULT_PAGE_TABLE_SIZE: usize = 0x800000;
+#[cfg(not(feature = "coverage"))]
 pub const DEFAULT_DMA_SIZE: usize = 0x100000;
+#[cfg(feature = "coverage")]
+pub const DEFAULT_DMA_SIZE: usize = 0x300000;
```
jyao1, line 11: can we use a dedicated DEFAULT_DMA_PROFILE_SIZE ?

PR #517, opened 2023-02-15
Extend `td-shim-layout-builder` tool and merge it into `td-layout-config`

gaojiaqi7:


devtools/td-layout-config/src/template/memory.jinja
```diff
@@ -0,0 +1,34 @@
...
+{{true|format_layout_border}} <- {{m.region.end|format_hex}}
+|{{m.name | format_name}}|   ({{m.region.end - m.region.start|format_hex}}) {{m.region.end - m.region.start|filesizeformat}}
+{%- endfor %}
+{{true|format_layout_border}} <- {{memory_regions_base|format_hex}}
+Total memory: {{total_memory_length|format_hex}} ({{total_memory_length|filesizeformat}})
+*/
+
+// Runtime Layout Configuration
+{%- for m in memory_regions %}
+{%- if m.name != "FREE" %}
```
xiaoyuxlu, line 19: why we need to hardcode the `FREE` region?

devtools/td-layout-config/src/template/memory.jinja
```diff
@@ -0,0 +1,34 @@
...
+{%- if m.tolm == false %}
+pub const {{m.name}}_BASE: usize = {{m.region.start | format_hex }};
+{%- endif %}
+pub const {{m.name}}_SIZE: usize = {{m.region.end - m.region.start | format_hex }}; // {{m.region.end - m.region.start|filesizeformat}}
+{%- endif %}
+{%- endfor %}
+
```
xiaoyuxlu, line 26: Can you calculate the total memory size to const TOTAL_MEMORY_SIZE  ?

xiaoyuxlu (changes requested):
Great! This will make it easier to generate layouts(runtime.rs and build_time.rs).

I have some inline comments.

devtools/td-layout-config/src/template/memory.jinja
gaojiaqi7: Removed, add a special entry type to filter out the free region instead of using name

devtools/td-layout-config/src/template/memory.jinja
gaojiaqi7: Sure, add `Total Usage` to indicate the total memory usage of the configuration

xiaoyuxlu (approved):

PR #521, opened 2023-02-20
td-shim: replace the `kernel` with `payload` in definitions

gaojiaqi7:


gaojiaqi7:
Fix: https://github.com/confidential-containers/td-shim/issues/460

jyao1 (approved):

PR #527, opened 2023-02-22
td-shim: use `payload_parameter` slice for `boot_kernel()`

gaojiaqi7:
The address of payload paramter may be different in different configurations so use slice instead.


jyao1 (approved):

jyao1:
conflict

gaojiaqi7:
resolved

PR #529, opened 2023-02-23
tdx-tdcall: use shared GPA for `GetQuote`

gaojiaqi7:
GHCI takes shared GPA as input for TDG.VP.VMCALL<GetQuote> leaf function.

To be compatible with different versions of qemu, add an option to choose whether to use shared GPA.


jyao1:
Question: is there any use case that the GPA is private?

gaojiaqi7:
It needs to be shared GPA according to GHCI

jyao1 (approved):

PR #530, opened 2023-02-28
td-shim-tools: use argument to determine the type of payload

gaojiaqi7:
Fix: https://github.com/confidential-containers/td-shim/issues/523

td-shim-tool needs the type of payload to generate metadta for the image.

Previously, we used feature to distinguish types of payload, its disadvantage
is that we need to compile different executable files for different types.

### DIFF
--- a/td-shim-tools/src/bin/td-shim-ld/main.rs
+++ b/td-shim-tools/src/bin/td-shim-ld/main.rs
@@ -9,9 +9,9 @@ extern crate clap;
 use std::io;
 use std::str::FromStr;
 
-use clap::ArgAction;
-use log::LevelFilter;
-use td_shim_tools::linker::TdShimLinker;
+use clap::{builder::PossibleValue, ArgAction};
+use log::{error, LevelFilter};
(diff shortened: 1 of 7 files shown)


td-shim-tools/src/bin/td-shim-ld/main.rs
```diff
@@ -37,6 +37,12 @@ fn main() -> io::Result<()> {
                 .takes_value(true)
                 .allow_invalid_utf8(false),
         )
+        .arg(
+            arg!(-t --"payload-type" "Payload Type: [linux, executable]")
```
jyao1, line 41: please add more description on what it is.

gaojiaqi7: Done.
```
-t, --payload-type <payload-type>
          Type of payload to be launched by td-shim
          
          [default: linux]

          Possible values:
          - linux:      Payload Binary is bzImage or vmlinux, follow Linux boot protocol
          - executable: Payload Binary is a PE/COFF or ELF executable image as payload
```

jyao1 (approved):

Issue #544, opened 2023-03-15
Warning reports by MIRAI 


liuw1:
Checking cc-measurement v0.1.0 (/home/liuwei/td-shim/cc-measurement)
warning: possible attempt to add with overflow
   --> cc-measurement/src/log.rs:136:35
    |
136 |             self.write_data(data, event_offset + data_offset);
    |                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^

warning: assertion failed
   --> cc-measurement/src/log.rs:211:9
    |
211 |         assert_eq!(digest.as_slice().len(), SHA384_DIGEST_SIZE);
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

Checking td-paging v0.1.0 (/home/liuwei/td-shim/td-paging)
warning: Invalid address 0x{} to BMFrameAllocator::reserve()
   --> td-paging/src/lib.rs:36:5
    |
36  |     frame::FRAME_ALLOCATOR.lock().reserve(addr);
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
note: related location
   --> td-paging/src/frame.rs:114:13
    |
114 | /             panic!(
115 | |                 "Invalid address 0x{:x} to BMFrameAllocator::reserve()",
116 | |                 addr
117 | |             );
    | |_____________^

warning: possible attempt to subtract with overflow
   --> td-paging/src/page_table.rs:163:9
    |
163 |         size -= page_size as i64;
    |         ^^^^^^^^^^^^^^^^^^^^^^^^

warning: possible index out of bounds
  --> td-payload/src/arch/x86_64/idt.rs:28:5
   |
28 |     idt_entries[vector as usize].set_ist(index);
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^

warning: possible attempt to add with overflow
   --> td-payload/src/arch/x86_64/paging.rs:198:32
    |
198 |             va = VirtAddr::new(va.as_u64() + frame.size())
    |                                ^^^^^^^^^^^^^^^^^^^^^^^^^^

warning: possible attempt to add with overflow
   --> td-payload/src/arch/x86_64/paging.rs:213:32
    |
213 |             va = VirtAddr::new(va.as_u64() + frame.size())
    |                                ^^^^^^^^^^^^^^^^^^^^^^^^^^

PR #545, opened 2023-03-20
Fix warnings reports by MIRAI

liuw1:
Fix: #544


td-payload/src/arch/x86_64/idt.rs
```diff
@@ -25,7 +25,11 @@ pub fn register(vector: u8, func: unsafe extern "C" fn()) {
 pub unsafe fn idt_set_ist(vector: u8, index: u8) {
     let idtr = store_idtr();
     let idt_entries = read_idt(&idtr);
-    idt_entries[vector as usize].set_ist(index);
+    if usize::from(vector) < idt_entries.len() {
+        idt_entries[vector as usize].set_ist(index);
+    } else {
+        log::info!("Set idt ist fail");
```
jyao1, line 31: I think it is fatal error.

liuw1: Resolved

jyao1 (approved):

Issue #556, opened 2023-04-06
Shared bit in the GPAs passed in R12 and R13 to TDG.VP.VMCALL<Service> is not set


kltseng:
The command and response arguments passed to tdvmcall_service() function are used as is in R12 and R13 when calling td_vmcall [1]. The calling function of tdvmcall_service() have those memories allocated from the calls to DmaMemory::new(1), e.g. MigrationSession::query() [2]. The address returned from DmaMemory::new(1) does not have shared bit set. This is not correct as the GHCI spec [3] which calls out R12 and R13 are shared buffer (Table 3-42: TDG.VP.VMCALL< Service >-Input Operands) which means the shared bit should be 1 of the addresses in R12 and R13

This failed a sanity check code in our VMM (VMware hypervisor) when handling the TDG.VP.VMCALL<Service> which checks if the R12 and R13 have shared bit set. 

If this is correct understanding, suggest to update the code as below:

diff --git a/tdx-tdcall/src/tdx.rs b/tdx-tdcall/src/tdx.rs
index bbdf66b..4de1ee4 100644
--- a/tdx-tdcall/src/tdx.rs
+++ b/tdx-tdcall/src/tdx.rs
@@ -411,8 +411,8 @@ pub fn tdvmcall_service(
     interrupt: u64,
     wait_time: u64,
 ) -> Result<(), TdVmcallError> {
-    let command = command.as_ptr() as u64;
-    let response = response.as_mut_ptr() as u64;
+    let command = command.as_ptr() as u64 | *SHARED_MASK;
+    let response = response.as_mut_ptr() as u64 | *SHARED_MASK;

[1] https://github.com/confidential-containers/td-shim/blob/5fa81efa1e43c8005d6d8099b5c099125c17b4b0/tdx-tdcall/src/tdx.rs#L427-L436
[2] https://github.com/intel-innersource/firmware.security.tdx.migtd.td/blob/main/migtd-core/src/migration/session.rs#L66-L67
[3] https://cdrdv2.intel.com/v1/dl/getContent/726792

PR #557, opened 2023-04-07
tdx-tdcall: use shared GPA for `TDG.VP.VMCALL<Service>`

gaojiaqi7:
Address of command and response buffer should be shared GPA according to GHCI.

Fix: https://github.com/confidential-containers/td-shim/issues/556


kltseng (commented):
This looks good to me and should address the issue. Thanks, Jiaqi.

jyao1 (approved):

PR #559, opened 2023-04-12
Add function to perform `sti;halt` operation

gaojiaqi7:
Fix: https://github.com/confidential-containers/td-shim/issues/561
Executing `hlt` instruction will cause a #VE to emulate the instruction. Safe halt operation `sti;hlt` which typically used for idle is not working in this case since `hlt` instruction must be the instruction next to `sti`. To use safe halt, `sti` must be executed just before `tdcall` instruction.

### DIFF
--- a/td-payload/src/arch/x86_64/apic.rs
+++ b/td-payload/src/arch/x86_64/apic.rs
@@ -27,6 +27,9 @@ pub fn enable_apic_interrupt() {
 }
 
 pub fn enable_and_hlt() {
+    #[cfg(feature = "tdx")]
+    tdx_tdcall::tdx::tdvmcall_sti_halt();
+    #[cfg(not(feature = "tdx"))]
     x86_64::instructions::interrupts::enable_and_hlt()
 }
 
--- a/tdx-tdcall/src/asm/mod.rs
+++ b/tdx-tdcall/src/asm/mod.rs
@@ -18,5 +18,5 @@ global_asm!(include_str!("tdvmcall.asm"));
 
 extern "win64" {
     pub(crate) fn asm_td_call(args: *mut c_void) -> u64;
-    pub(crate) fn asm_td_vmcall(args: *mut c_void) -> u64;
+    pub(crate) fn asm_td_vmcall(args: *mut c_void, do_sti: u64) -> u64;
 }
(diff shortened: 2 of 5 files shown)


xiaoyuxlu:
I prefer to add a tdcall_halt separately.

gaojiaqi7:
I think we should reuse existing code as much as possible

xiaoyuxlu (approved):
Since the kernel does the same, it works fine for me.

gaojiaqi7:
This PR uses the same approach as the Kernel 6.2 which issues an `sti` before the `tdcall`.

Kernel 6.3 remove the `ISSUE_STI` approach due to the specfic `idle` implementaion of kernel, it is not specific to TDX, so I think we can use the `ISSUE_STI` approach in our implementation for now.

gaojiaqi7:
TDG.VP.VMCALL<Instruction.HLT>

The TD is expected to clear this flag if RFLAGS.IF == 1 or the TDCALL instruction (that invoked TDG.VP.TDVMCALL(Instruction.HLT))  immediately follows an STI instruction, otherwise this flag should be set.

td-payload/src/arch/x86_64/apic.rs
```diff
@@ -27,6 +27,9 @@ pub fn enable_apic_interrupt() {
 }
 
 pub fn enable_and_hlt() {
+    #[cfg(feature = "tdx")]
+    tdx_tdcall::tdx::tdvmcall_safe_halt();
```
jyao1, line 31: please rename sti_halt().

tdx-tdcall/src/asm/mod.rs
```diff
@@ -18,5 +18,5 @@ global_asm!(include_str!("tdvmcall.asm"));
 
 extern "win64" {
     pub(crate) fn asm_td_call(args: *mut c_void) -> u64;
-    pub(crate) fn asm_td_vmcall(args: *mut c_void) -> u64;
+    pub(crate) fn asm_td_vmcall(args: *mut c_void, do_sti: u64) -> u64;
```
jyao1, line 21: can we keep existing api, and add td_vmcall_ex?

gaojiaqi7: This is not a public function, added a `td_vmcall_ex` below

td-payload/src/arch/x86_64/apic.rs
gaojiaqi7: done, renamed to tdvmcall_sti_halt

jyao1 (approved):

Issue #558, opened 2023-04-11
R12 of `TDG.VP.VMCALL<Instruction.HLT>` needs to be taken into account


gaojiaqi7:
https://github.com/confidential-containers/td-shim/blob/main/tdx-tdcall/src/tdx.rs#L77-L87

PR #560, opened 2023-04-12
tdx-tdcall: use interrupt blocked flag for TDVMCALL HLT

gaojiaqi7:
Fix: https://github.com/confidential-containers/td-shim/issues/558
According to GHCI, interrupt blocked flag should be cleared if `RFLAGS.IF=1` or the `tdcall` instruction immediately follows an `sti` instruction, otherwise this flag should be set


xiaoyuxlu (approved):
looks good.

PR #566, opened 2023-05-31
td-shim-tools/tee-info-hash: correct the value parser of `out_bin`

gaojiaqi7:


jyao1 (approved):

jyao1:
ci server is down. Not sure when it can recover.

PR #567, opened 2023-05-31
td-shim-tools/tee-info-hash: add option of string format output

gaojiaqi7:
https://github.com/confidential-containers/td-shim/issues/568

Add an option to output the hash of tee info to stdout in string format

### DIFF
--- a/td-shim-tools/src/bin/td-shim-tee-info-hash/main.rs
+++ b/td-shim-tools/src/bin/td-shim-tee-info-hash/main.rs
@@ -145,5 +145,12 @@ fn main() -> io::Result<()> {
         &config.output.display()
     );
 
+    println!(
+        "{}",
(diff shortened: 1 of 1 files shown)


td-shim-tools/src/bin/td-shim-tee-info-hash/main.rs
```diff
@@ -26,6 +26,8 @@ struct Config {
     pub image: String,
     // Output binary of tee info hash
     pub output: PathBuf,
+    // Ouput tee info hash to stdout
+    pub print_flag: bool,
```
jyao1, line 30: I think we can unconditionally print it. NO need to add flag.

jyao1 (approved):

jyao1:
ci server is down. Not sure when it can recover.

PR #576, opened 2023-06-25
Change the type for heap_start to u64.

longlongyang:
The reason is that the target platform is always 64 bit.


xiaoyuxlu (approved):
Per discussion, the current platform is 64-bit. So use u64.

Issue #596, opened 2023-09-12
MADT MP Wakeup structure: mailbox version should be `0`? 


dimakuv:
TD-Shim constructs a MADT MP Wakeup structure that contains the mailbox version `1`: https://github.com/confidential-containers/td-shim/blob/34b31e37816b3f9af159c218a9fe789588670cfd/td-shim/src/bin/td-shim/mp.rs#L87

However, when I read the current TDX/ACPI specifications, I see that the mailbox version is supposed to be `0`:
- https://uefi.org/sites/default/files/resources/ACPI_Spec_6_5_Aug29.pdf, Release 6.5 from Aug 29, 2022, in Table 5.43 "Multiprocessor Wakeup Structure"
- https://www.intel.com/content/dam/develop/external/us/en/documents/intel-tdx-guest-hypervisor-communication-interface.pdf, version from September 2020, Table 4 on page 37.

Am I just using the old specs? Where can I find the latest specs that have the mailbox version of `1`?

jyao1:
Yes, it should be 0. https://uefi.org/specs/ACPI/6.5/05_ACPI_Software_Programming_Model.html#multiprocessor-wakeup-structure

PR #597, opened 2023-09-13
td-shim: correct MPWK mailbox version

gaojiaqi7:
According to the latest ACPI spec, the mailbox version of MPWK should be 0.

Fix https://github.com/confidential-containers/td-shim/issues/596


dimakuv (approved):
LGTM

jyao1 (approved):

Issue #604, opened 2023-09-22
Cannot reproduce binary in different paths when `secure-boot` feature is enabled


gaojiaqi7:
More specifically, when `ring` is used as a dependency, the binaries compiled in different paths will be different.

Try:
```
cargo image --release --features secure-boot
```

PR #605, opened 2023-09-26
add `Dockerfile` for reproducible build

gaojiaqi7:
The reproducibility of td-shim is limited by the same system user and source code path, so we can use a Docker container to reproduce binaries.

Resolves: https://github.com/confidential-containers/td-shim/issues/604

### DIFF
--- a/devtools/dev_container/Dockerfile
+++ b/devtools/dev_container/Dockerfile
@@ -37,3 +37,5 @@ RUN set -eux; \
     git clone https://github.com/sslab-gatech/Rudra.git; \
     cd Rudra; \
     ./install-release.sh;
+
+RUN git clone https://github.com/confidential-containers/td-shim.git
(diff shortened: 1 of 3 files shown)


jyao1:
Would you please reuse https://github.com/confidential-containers/td-shim/blob/main/devtools/dev_container/Dockerfile, as much as possible?

devtools/dev_container/Dockerfile
```diff
@@ -37,3 +37,6 @@ RUN set -eux; \
     git clone https://github.com/sslab-gatech/Rudra.git; \
     cd Rudra; \
     ./install-release.sh;
+
+RUN git clone https://github.com/confidential-containers/td-shim.git
```
jyao1, line 41: can we move this to docker.sh?

jyao1: @jiangliu , are you OK with that change?

jyao1 (approved):

PR #606, opened 2023-10-08
Add design-for-payload document.

jyao1:


jyao1:
@jiangliu , @sameo would you please review this change?

Issue #617, opened 2023-10-31
ELF payload: what goes into MRTD and RTMRs?


dimakuv:
I have my ELF payload. I link it with TD-Shim firmware like this:
```bash
$ cargo run -p td-shim-tools --bin td-shim-ld -- \
  target/x86_64-unknown-none/release/ResetVector.bin target/x86_64-unknown-none/release/td-shim \
  -t executable -p my-payload \
   -o resulting-binary
```

The resulting binary is 16MB in size:
```bash
$ $ ls -lah resulting-binary
-rw-r--r-- 16M Oct 31 01:27 resulting-binary
```

Then this resulting binary is fed to QEMU to start a TDX VM:
```bash
qemu ... -bios resulting-binary
```

I'm trying to make sense of the MRTD and RTMR measurements that I observe:

- My original payload:
```
  MRTD    = d0 78 10 95 8e 42 82 14 ...
  RTMR[0] = 2e 8f f9 2e d0 0d ed bc ...
  RTMR[1] = d7 c0 29 9f bf d0 a9 7d ...
  RTMR[2] = 00 00 00 00 00 00 00 00 ...
  RTMR[3] = 00 00 00 00 00 00 00 00 ...
```

- My slightly modified payload (one byte difference):
```
  report MRTD    = 52 cb 42 08 f1 15 3a d5 ...
  report RTMR[0] = 2e 8f f9 2e d0 0d ed bc ...
  report RTMR[1] = 3b 1a d8 a8 a7 61 8a c4 ...
  report RTMR[2] = 00 00 00 00 00 00 00 00 ...
  report RTMR[3] = 00 00 00 00 00 00 00 00 ...
```

I understand why RTMRs 2 and 3 are all-zeros. VMM, TD-Shim, and my payload do **not** touch these RTMRs.

I think I understand why `RTMR[0]` is a constant -- it's because I run QEMU with the same parameters (like the list of exposed devices), so `RTMR[0]` reflects the VMM inputs (the TD Hob generated by VMM and handed over to TD-Shim).

But I'm confused with `MRTD` and `RTMR[1]`:
- Apparently, MRTD reflects **both** the TD-Shim firmware itself and my payload binary.
- `RTMR[1]` reflects Payload and PayloadParam? But the payload binary was already measured into MRTD; why is it also measured into `RTMR[1]`? Why is `RTMR[1]` changed in my experiment?
[...]

PR #618, opened 2023-11-03
Measure payload only when EXTENDMR is not set

gaojiaqi7:
Fixes https://github.com/confidential-containers/td-shim/issues/617

### DIFF
--- a/td-shim/src/bin/td-shim/main.rs
+++ b/td-shim/src/bin/td-shim/main.rs
@@ -146,6 +146,7 @@ pub extern "win64" fn _start(
     // If the Payload Information GUID HOB is present, try to boot the Linux kernel.
     if let Some(payload_info) = dynamic_info.payload_info {
         boot_linux_kernel(
+            &static_info,
             &payload_info,
             &dynamic_info.acpi_tables,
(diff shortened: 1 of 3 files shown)


td-shim-tools/src/metadata.rs
```diff
@@ -234,7 +234,7 @@ pub fn default_metadata_sections(payload_type: PayloadType) -> MetadataSections
                 memory_address: TD_SHIM_PAYLOAD_BASE as u64,
                 memory_data_size: TD_SHIM_PAYLOAD_SIZE as u64,
                 r#type: TDX_METADATA_SECTION_TYPE_PAYLOAD,
-                attributes: 0,
+                attributes: 3,
```
jyao1, line 237: why need to change this?

dimakuv (changes requested):
The PR doesn't work for me. (Maybe I'm doing something wrong?)

Please see https://github.com/confidential-containers/td-shim/issues/617#issuecomment-1794328608

dimakuv (approved):
The latest changes work fine, thanks!

jyao1 (approved):

PR #631, opened 2023-12-07
add memory address/size check for TD_INFO.

### LINKED ISSUE #629: Need to check memory_address and memory_data_size of td_info section in metadata
memory_address and memory_data_size of td_info section should be zero.

longlongyang:
fix #629


jyao1 (approved):

jyao1 (changes requested):
Should we add check for data_offset and raw_data_size to ensure they are covered by BFV?

longlongyang:
> Should we add check for data_offset and raw_data_size to ensure they are covered by BFV?

This check is already there:
https://github.com/confidential-containers/td-shim/blob/850bb35490f634041c6bd049e55de6523613c7d6/td-shim/src/metadata.rs#L425-L430

jyao1 (approved):

PR #635, opened 2023-12-14
upgrade dependency `env_logger`

gaojiaqi7:
To unuse the crate `atty` which has a potential security issue and is unmaintained.


Cargo.lock
```diff
@@ -421,12 +421,12 @@ checksum = "a26ae43d7bcc3b814de94796a5e736d4029efb0ee900c12e2d54c993ad1a1e07"
 
 [[package]]
 name = "env_logger"
-version = "0.9.3"
+version = "0.10.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a12e6657c4c97ebab115a42dcee77225f7f482cdd841cf7088c657a42e9e00e7"
+checksum = "95b3f3e67048839cb0d0781f445682a35113da7121f7c949db0e2be96a4fbece"
 dependencies = [
- "atty",
```
jyao1, line 428: I think we also need to remove `atty` related entry.

gaojiaqi7: Removed by upgrading `clap` to `4.0`

jyao1 (approved):

Issue #632, opened 2023-12-11
Request to support to locate TDX metadata for TDVF with td-shim-tee-info-hash tool


jiazhang0:
Currently, td-shim follows Intel TDX Virtual Firmware Design Guide (Section 11.1) to locate TDX metadata, while TDVF follows OVMF GUIDed Structure to locate TDX metadata.

Actually, TDVF uses the same TDX metadata with td-shim but they just have different method to locate it. So it makes sense to implement the locating method for TDVF in td-shim-tee-info-hash tool rather than writing a new tool for TDVF.

OuyangHang33:
I will check this one

OuyangHang33:
td-shim-tee-info-hash tool can support 2 locating method mentioned above, but it set metadata buffer size based on TD_SHIM_METADATA_SIZE, which is 0x1000. 
I get the length from TDVF descriptor structure and set the metadata buffer based on it. Now it can locate TDX metadata correctly.

PR #651, opened 2024-01-16
Get metadata buffer size from TDVF DESCRIPTOR

OuyangHang33:
Fix: #632

### DIFF
--- a/td-shim-tools/src/tee_info_hash.rs
+++ b/td-shim-tools/src/tee_info_hash.rs
@@ -14,7 +14,6 @@ use std::io::Read;
 use std::io::Seek;
 use std::io::SeekFrom;
 use std::mem::size_of;
-use td_layout::build_time::TD_SHIM_METADATA_SIZE;
 use td_shim::metadata::*;
 use zeroize::Zeroize;
 
@@ -243,11 +242,9 @@ impl TdInfoStruct {
             .seek(SeekFrom::Start(metadata_off as u64))
             .unwrap();
 
-        let mut metadata_buf = [0; TD_SHIM_METADATA_SIZE as usize];
-
-        raw_image_file.read_exact(&mut metadata_buf).unwrap();
-
-        let desc = &metadata_buf[size_of::<TdxMetadataGuid>()..];
+        let mut desc_buf = [0; size_of::<TdxMetadataGuid>() + size_of::<TdxMetadataDescriptor>()];
+        raw_image_file.read_exact(&mut desc_buf).unwrap();
+        let desc = &desc_buf[size_of::<TdxMetadataGuid>()..];
 
         // Signature	        0	CHAR8[4]	    4	'TDVF' signature
         // Length	            4	UINT32	        4	Size of the structure (d)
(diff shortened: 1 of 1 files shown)


jyao1:
@jiazhang0 , would you please review this patch?

jyao1:
@OuyangHang33 , please describe what test you have done.

OuyangHang33:
I test this tool with OVMF image and td-shim image, and can calculate MRTD and RTMR values successfully.

OuyangHang33:
@jiazhang0 could you help me review this patch? Thanks!

jyao1 (approved):

td-shim-tools/src/tee_info_hash.rs
```diff
@@ -243,11 +242,9 @@ impl TdInfoStruct {
...
             .unwrap();
 
-        let mut metadata_buf = [0; TD_SHIM_METADATA_SIZE as usize];
-
-        raw_image_file.read_exact(&mut metadata_buf).unwrap();
-
-        let desc = &metadata_buf[size_of::<TdxMetadataGuid>()..];
+        let mut desc_buf = [0; size_of::<TdxMetadataGuid>() + size_of::<TdxMetadataDescriptor>()];
+        raw_image_file.read_exact(&mut desc_buf).unwrap();
+        let desc = &desc_buf[size_of::<TdxMetadataGuid>()..];
```
Xynnn007, line 247: I am a little confused. The logic here means
1. The memory starts from `metadata_off` is a `TdxMetadataGuid` and a `TdxMetadataDescriptor`
2. read the `TdxMetadataGuid` and `TdxMetadataDescriptor`
3. `desc` is assigned with the memory slice of `TdxMetadataDescriptor`

td-shim-tools/src/tee_info_hash.rs
```diff
@@ -261,6 +258,15 @@ impl TdInfoStruct {
...
 
+        raw_image_file
+            .seek(SeekFrom::Start(
+                metadata_off as u64 + size_of::<TdxMetadataGuid>() as u64,
+            ))
+            .unwrap();
+        let mut metadata_buf = vec![0; descriptor.length as usize];
+        raw_image_file.read_exact(&mut metadata_buf).unwrap();
+        let desc = &metadata_buf[0..];
+
```
Xynnn007, line 269: Continuing from above. Then we do
1. read `TdxMetadataDescriptor` and assign the value to variable `desc`

So, seems like the PR does nothing?

td-shim-tools/src/tee_info_hash.rs
OuyangHang33: The previous code assumed the buffer size is TD_SHIM_METADATA_SIZE which is 0x1000. So it read from metadata_off to metadata_off+0x1000. 
But actually it should read TdxMetadataDescriptor first and get the length from TdxMetadataDescriptor structure and set is as buffer size.

td-shim-tools/src/tee_info_hash.rs
OuyangHang33: Previous code can't support OVMF image, it's metadata size is smaller than `0x1000`.
I get the length from TDVF descriptor structure and set the metadata buffer based on it. Now it can locate TDX metadata for OVMF image correctly.

Xynnn007 (approved):
Thanks @OuyangHang33 for the explaination. I test locally and this surely fix the bug.

Issue #654, opened 2024-02-01
Publish td-shim in crate.io 


NK-fengsf:
I am writing to ask  if there are plans to release an official crate for td-shim in crate.io? 

Currently, td-shim is still in development, but it can already meet the basic usage needs. I am considering using td-shim in dragonball-sandbox , which  is a collection of Rust crates to help build custom Virtual Machine Monitors and used by kata-containers.

The data structures and functions required for td-shim by dragonball-sandbox include：
1.  The data structures of td-shim metadata and related crate， such as TdxMetadataDescriptor and TdxMetadataSection. 
2.  The data structure of TD HOB and related crate, such as td_shim::{TD_ACPI_TABLE_HOB_GUID, TD_E820_TABLE_HOB_GUID, TD_PAYLOAD_INFO_HOB_GUID};   td_shim::PayloadInfo,  td-shim::acpi, td_uefi_pi::pi::hob; td_uefi_pi::{hob, pi, pi::guid};

If there are plans , please let me know the specific release time and features.

Thank you!

NK-fengsf:
@gaojiaqi7  @jyao1

jyao1:
I think we just need to post the td-shim metadata related crate to crate.io, in order to let VMM consume it, right?

NK-fengsf:
> I think we just need to post the td-shim metadata related crate to crate.io, in order to let VMM consume it, right?
We need  data structure definitions of  both the TD-shim metadata and TD HOB.   The td-shim parse function in td-shim-tools crate are also needed.

studychao:
@jyao1 hi, we have updated the issue description to include all the crate we need in td-shim, please take a look.

OuyangHang33:
I will check this issue

studychao:
Hi @jyao1 @OuyangHang33 , although this issue is closed, but I couldn't find related crates in the `crate.io`. I have tried to search "td-shim" or "td-shim-interface" but I found nothing.

What's the name for the crate that is published? Thanks!

OuyangHang33:
Hi @studychao , we plan to publish crate after #657 tested and merged. I think we need reopen it.
Hi @jyao1 , please correct me if I'm wrong.

OuyangHang33:
td-shim-interface url: https://crates.io/crates/td-shim-interface

PR #660, opened 2024-02-29
Split file operation from loader.rs and change parse() API

OuyangHang33:
Fix: #654

Splited the file operation and std env from loader.rs. Moved the read file operation to read_file.rs.

Test the final.bin with td-shim-checker tool and it can get the right result.


NK-fengsf:
LGFM

studychao (approved):
LGTM

jyao1 (approved):

PR #661, opened 2024-03-03
td-payload: add shadow pages for shared memory

gaojiaqi7:
Data in shared memory provided by VMM is untrusted so from a security perspective the data shall be consumed after being copied into private memory.

Private shadow for `SharedMemory` is provided to add a more easy and functional safe mechanism to use the untrusted data. It is 1:1 allocated for a shared memory request, and users can use the data with `copy_to_private_shadow` directly to get a private copy with additional memory allocation/free operations.

### DIFF
--- a/td-payload/src/mm/shared.rs
+++ b/td-payload/src/mm/shared.rs
@@ -4,38 +4,55 @@
 
 use core::{alloc::Layout, ptr::NonNull};
 use linked_list_allocator::LockedHeap;
+use spin::Once;
 
 use super::SIZE_4K;
 use crate::arch::shared::decrypt;
 
 static SHARED_MEMORY_ALLOCATOR: LockedHeap = LockedHeap::empty();
+static SHADOW_OFFSET: Once<usize> = Once::new();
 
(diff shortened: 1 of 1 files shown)


td-payload/src/mm/shared.rs
```diff
@@ -9,31 +9,52 @@ use super::SIZE_4K;
...
+
     // Set the shared memory region to be shared
-    decrypt(start, size);
+    decrypt(shared_start, shared_size);
     // Initialize the shared memory allocator
     unsafe {
-        SHARED_MEMORY_ALLOCATOR.lock().init(start as *mut u8, size);
+        SHARED_MEMORY_ALLOCATOR
+            .lock()
+            .init(shared_start as *mut u8, shared_size);
```
jyao1, line 25: If we use double allocators, then we assume that the allocator is deterministic, not randomized. Right?

I am not sure why we need 2nd allocator, why not just use "address + offset" for the shadow?

gaojiaqi7: Got it, will follow this way

jyao1 (approved):

Issue #662, opened 2024-03-03
td-payload: issues with `SharedMemory`


gaojiaqi7:
- [ ] `size` of `new` method should be checked for page size aligned
- [ ] unwrap and panic should be removed in free functions
- [ ] `LockedHeap` uses the memory for allocation to maintain the allocation status, but it will expose the status if the whole shared memory is decrypted in `init_shared_memory`

PR #664, opened 2024-03-03
td-payload: decrypt private memory during allocation

gaojiaqi7:
Decrypt the private memory at initialization will leak the data of the linked list allocator.

The solution can be moving decrytion to the moment the shared pages are allocated and encypting the shared memory before they are freed.

Closes: https://github.com/confidential-containers/td-shim/issues/662

### DIFF
--- a/td-payload/src/mm/shared.rs
+++ b/td-payload/src/mm/shared.rs
@@ -6,13 +6,11 @@ use core::{alloc::Layout, ptr::NonNull};
 use linked_list_allocator::LockedHeap;
 
 use super::SIZE_4K;
-use crate::arch::shared::decrypt;
+use crate::arch::shared::{decrypt, encrypt};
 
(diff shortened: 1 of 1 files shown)


jyao1:
I am not sure why we need this.

The expectation is that we pre-allocate a chunk of shared memory.
There should be NO secrete in the shared memory. Why we need to encrypt?

gaojiaqi7:
> I am not sure why we need this.
> 
> The expectation is that we pre-allocate a chunk of shared memory. There should be NO secrete in the shared memory. Why we need to encrypt?

The allocator `LockedHeap` uses a linked list which puts allocation information in the free memory blocks.

Another solution can be using/realizing a bitmap allocator to put all the allocation information in the private memory

gaojiaqi7:
Close as the allocator data of shared memory is not confidence

Issue #667, opened 2024-03-10
Fuzzing test cases build fail in td-shim-interface/src


liuw1:
```~/work/td-shim/td-shim/td-shim-interface/src ~/work/td-shim/td-shim
Error: could not read the manifest file: /home/runner/work/td-shim/td-shim/td-shim-interface/fuzz/Cargo.toml

Caused by:
    No such file or directory (os error 2)
```
@OuyangHang33 Could you please help to fix it?

jyao1:
why CI does not catch such issue?

liuw1:
> why CI does not catch such issue?

The fuzzing script doesn't handle this scenario.

OuyangHang33:
I will check it

PR #669, opened 2024-03-11
Fuzzing: Enhance test scripts to detect issue #667

liuw1:
Fix #668

PR #670, opened 2024-03-11
Move td-uefi-pi fuzz folder

OuyangHang33:
Fix: #667 

Rename and move td-uefi-pi fuzz folder to td-shim-interface.


OuyangHang33:
After #669 merged, this PR is needed rerun CI again and get pass.

Issue #665, opened 2024-03-05
max payload size?


mythi:
I tried to build an image with a systemd Unified Kernel Image as the payload but the build asserts:

```
assertion failed: data <= 0xffffff
stack backtrace:
   0: rust_begin_unwind
             at /rustc/2a3e63551fe21458637480a97b65a2d15dec8062/library/std/src/panicking.rs:645:5
   1: core::panicking::panic_fmt
             at /rustc/2a3e63551fe21458637480a97b65a2d15dec8062/library/core/src/panicking.rs:72:14
   2: core::panicking::panic
             at /rustc/2a3e63551fe21458637480a97b65a2d15dec8062/library/core/src/panicking.rs:144:5
   3: td_shim::write_u24
             at ./td-shim/src/lib.rs:105:5
   4: td_shim_tools::linker::FvHeaderByte::build_tdx_payload_fv_header
             at ./td-shim-tools/src/linker.rs:119:9
   5: td_shim_tools::linker::TdShimLinker::build
             at ./td-shim-tools/src/linker.rs:367:34
   6: td_shim_ld::main
             at ./td-shim-tools/src/bin/td-shim-ld/main.rs:93:5
   7: core::ops::function::FnOnce::call_once
             at /rustc/2a3e63551fe21458637480a97b65a2d15dec8062/library/core/src/ops/function.rs:250:5
```

AFAIUI, the spec does not limit the size but says (for firmware files section): "If the section size is 0xFFFFFF then the size is defined by a 32-bit integer that follows the 32-bit section header.". Would it make sense to support this 'extended length' approach?

jyao1:
Technically, it is possible. 
May I know how big the UKI image is?

mythi:
a UKI image typically contains the distro's initrd or it can be a custom build with the necessary content. I can't say the upper limit but it's more than 16M. 

I had a UKI that was ~190M and had to make the following modifications to get the new image layout generated:

```
diff --git a/devtools/td-layout-config/config_image.json b/devtools/td-layout-config/config_image.json
index dbf4e39..391d927 100644
--- a/devtools/td-layout-config/config_image.json
+++ b/devtools/td-layout-config/config_image.json
@@ -4,7 +4,7 @@
     "TempStack": "0x020000",
     "TempHeap": "0x020000",
     "Metadata": "0x001000",
-    "Payload": "0xC2D000",
+    "Payload": "0xC2D0000",
     "Ipl": "0x348000",
[...]

PR #675, opened 2024-03-20
td-layout: load payload image into physical memory space

gaojiaqi7:
The old implementation loads the payload image into ROM space which limited the size of the payload to be smaller than 16MiB.

 - The previous layout 1:1 maps the image to the ROM space, as we move payload to physical memory space, ROM layout and image layout will be different in `td-layout-config` tool.
 - Firmware volume header has limitation on the size of the data inside which needs to be smaller than maximum value of u24 (0xffffff). To lift the restriction, FV header is removed for payload images as it is not mandatory.
 - At runtime, if td-shim image is not loaded into memory 1:1, the metadata offset at firmware_end - 0x20 will be invalid because it is the offset from the beginning of the image. The value inside the OVMF GUID table is the offset from the end of the image which is not changed after loading into memory.

Related issue: https://github.com/confidential-containers/td-shim/issues/665

### DIFF
--- a/td-layout/src/build_time.rs
+++ b/td-layout/src/build_time.rs
@@ -1,4 +1,4 @@
-// Copyright (c) 2021 - 2023  Intel Corporation
+// Copyright (c) 2021 - 2024  Intel Corporation
 //
 // SPDX-License-Identifier: BSD-2-Clause-Patent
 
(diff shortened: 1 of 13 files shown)


td-layout/src/build_time.rs
```diff
@@ -7,72 +7,64 @@
...
-|               TEMP_STACK               |   (0x20000) 128 KB
+|               TEMP_STACK               |   (0x20000) 128 kB
 +----------------------------------------+ <- 0x61000
-|               TEMP_HEAP                |   (0x20000) 128 KB
+|               TEMP_HEAP                |   (0x20000) 128 kB
 +----------------------------------------+ <- 0x81000
-|                  FREE                  |   (0x1000) 4 KB
-+----------------------------------------+ <- 0x82000
-|                PAYLOAD                 |   (0xC2D000) 12.18 MB
+|                  FREE                  |   (0x0) 0 B
```
jyao1, line 18: what is this?

td-layout/src/memslice.rs
```diff
@@ -39,7 +39,7 @@ impl SliceType {
         match self {
             SliceType::Config => "Config",
             SliceType::TdHob => "TdHob",
-            SliceType::ShimPayload => "ShimPayload",
+            SliceType::PayloadImage => "PayloadImage",
```
jyao1, line 42: why change this?

devtools/td-layout-config/config_image.json
```diff
@@ -4,7 +4,7 @@
     "TempStack": "0x020000",
     "TempHeap": "0x020000",
     "Metadata": "0x001000",
-    "Payload": "0xC2D000",
+    "Payload": "0xC2E000",
```
jyao1, line 7: why we need to enlarge for load into physical memory?

td-shim-interface/src/metadata.rs
```diff
@@ -18,6 +18,21 @@ pub const TDX_METADATA_GUID: Guid = Guid::from_fields(
...
+    0x45f7,
+    [0xba, 0xea, 0xa3, 0x66, 0xc5, 0x5a, 0x08, 0x2d],
+);
+
+pub const OVMF_TABLE_TDX_METADATA_GUID: Guid = Guid::from_fields(
+    0xe47a6535,
+    0x984a,
+    0x4798,
+    [0x86, 0x5e, 0x46, 0x85, 0xa7, 0xbf, 0x8e, 0xc2],
+);
```
jyao1, line 34: why we need this?

PR #679, opened 2024-03-30
td-layout: move payload to the start of image file

gaojiaqi7:
When the payload is loaded into physical memory rather than the ROM space, image-rom mapping of other regions will not be affected if payload is at the start of the image file.

### DIFF
--- a/devtools/td-layout-config/config_image.json
+++ b/devtools/td-layout-config/config_image.json
@@ -1,10 +1,10 @@
 {
+    "Payload": "0xC2D000",
     "Config": "0x040000",
     "Mailbox": "0x001000",
     "TempStack": "0x020000",
     "TempHeap": "0x020000",
     "Metadata": "0x001000",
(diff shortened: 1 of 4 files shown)


jyao1:
This is not what we have discussed.
I proposed to add large_payload section in front, and keep original layout as is.

gaojiaqi7:
> This is not what we have discussed. I proposed to add large_payload section in front, and keep original layout as is.

The reason is adding a new section bring lots of changes in implemetation:
1. build time: where to put the payload image - `payload` or `large_payload`?
2. build time: which payload section will be adding into metadata?
3. runtime: where td-shim can find the payload image?

For the metadata offset issue we have discussed, two approaches use the same way to locate the metadata: `metadata offset - sizeof payload section`.

So I think moving the payload section to the begining of image file will be a better approach in implementation. Do you have any concern on this?

PR #694, opened 2024-05-10
td shim: support large payload

gaojiaqi7:
- The metadata offset at the `end - 0x20`of image is the offset of the
   metadata from the beginning of the file, so subtract the large payload
   size when calculating metadata offset.

 - As upstream qemu does not support TDVF `payload` section type, try both
   `Payload` and `LargePayload` region to find the firmware volume that
   payload image may be packaged in.


td-layout/src/build_time.rs
```diff
@@ -7,72 +7,76 @@
 /*
 Image Layout
 +----------------------------------------+ <- 0x0
-|                 CONFIG                 |   (0x40000) 256 KB
+|             LARGE_PAYLOAD              |   (0x0) 0 B
++----------------------------------------+ <- 0x0
+|                 CONFIG                 |   (0x40000) 256 kB
 +----------------------------------------+ <- 0x40000
-|                MAILBOX                 |   (0x1000) 4 KB
+|                MAILBOX                 |   (0x1000) 4 kB
```
jyao1, line 14: why change KB to kB?

td-shim/src/bin/td-shim/shim_info.rs
```diff
@@ -35,7 +37,8 @@ impl BootTimeStatic {
     // Validate the metadata and get the basic infomation from
     // it if any
     pub fn new() -> Option<Self> {
-        let metadata_offset = unsafe { *((u32::MAX - TDX_METADATA_OFFSET + 1) as *const u32) };
+        let metadata_offset = unsafe { *((u32::MAX - TDX_METADATA_OFFSET + 1) as *const u32) }
+            - TD_SHIM_LARGE_PAYLOAD_SIZE;
```
jyao1, line 41: will this overflow?

gaojiaqi7: Added the `checked_sub` to prevent potential overflow, thanks

PR #695, opened 2024-05-10
0510/td shim tools support large payload

gaojiaqi7:
Use `FfsFileHeader2` and `CommonSectionHeader2` for payload FV.

If payload binary is larger than the `Payload` region size, then it
will be treated as a large payload and it will be packaged in the
start of the image and loaded into physical memory as indicated by
metadata.


td-layout/src/build_time.rs
```diff
@@ -7,72 +7,76 @@
 /*
 Image Layout
 +----------------------------------------+ <- 0x0
-|                 CONFIG                 |   (0x40000) 256 KB
+|             LARGE_PAYLOAD              |   (0x0) 0 B
++----------------------------------------+ <- 0x0
+|                 CONFIG                 |   (0x40000) 256 kB
 +----------------------------------------+ <- 0x40000
-|                MAILBOX                 |   (0x1000) 4 KB
+|                MAILBOX                 |   (0x1000) 4 kB
```
jyao1, line 14: why change KB to kB?

Issue #685, opened 2024-04-16
clean up vec usage 


jyao1:
For example, prepare_acpi_table().

PR #700, opened 2024-05-17
td-shim: remove the use of `Vec` in `e820`

gaojiaqi7:
Reducing the use of heap can improve the functional safety. For e820 implementation, use an fixed `MAX_E820_ENTRY` size array to replace the vector.

Relate issue: https://github.com/confidential-containers/td-shim/issues/685

### DIFF
--- a/td-shim/src/bin/td-shim/e820.rs
+++ b/td-shim/src/bin/td-shim/e820.rs
@@ -3,8 +3,6 @@
 // SPDX-License-Identifier: BSD-2-Clause-Patent
 
 use core::ptr::slice_from_raw_parts;
-
-use alloc::vec::Vec;
 use td_shim::e820::{E820Entry, E820Type};
 
 // Linux BootParam supports 128 e820 entries, so...
@@ -12,7 +10,8 @@ const MAX_E820_ENTRY: usize = 128;
 
 #[derive(Debug)]
 pub struct E820Table {
-    entries: Vec<E820Entry>,
+    entries: [E820Entry; MAX_E820_ENTRY],
+    num_entries: usize,
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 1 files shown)


td-shim/src/bin/td-shim/e820.rs
```diff
@@ -158,27 +154,58 @@ impl E820Table {
...
+    fn insert_entry(&mut self, index: usize, entry: E820Entry) -> Result<(), E820Error> {
+        if index >= MAX_E820_ENTRY {
+            return Err(E820Error::InvalidIndex);
+        }
+        if self.num_entries >= MAX_E820_ENTRY {
+            return Err(E820Error::TooManyEntries);
+        }
+
+        // Move all entries after the index back one position
+        for idx in (index + 1..self.num_entries + 1).rev() {
```
jyao1, line 183: what if the index > self.num_entries?

td-shim/src/bin/td-shim/e820.rs
```diff
@@ -158,27 +154,58 @@ impl E820Table {
...
+        Ok(())
+    }
+
+    fn remove_entry(&mut self, index: usize) -> Result<(), E820Error> {
+        if index >= MAX_E820_ENTRY {
+            return Err(E820Error::InvalidIndex);
+        }
+
+        // Move all entries after the index forward one position
+        for idx in index + 1..self.num_entries {
```
jyao1, line 198: what if the index > self.num_entries?

td-shim/src/bin/td-shim/e820.rs
```diff
@@ -158,27 +154,58 @@ impl E820Table {
...
+            self.entries[idx + 1] = self.entries[idx];
+        }
+
+        self.entries[index] = entry;
+        self.num_entries += 1;
+        Ok(())
+    }
+
+    fn remove_entry(&mut self, index: usize) -> Result<(), E820Error> {
+        if index >= self.num_entries {
```
jyao1, line 193: What if self.num_entries is 0 ?

gaojiaqi7: if `self.num_entries == 0`, the `index >= self.num_entries` will always be true, then it will return error

jyao1 (approved):

PR #701, opened 2024-05-17
td-shim: remove vector in `secure_boot`

gaojiaqi7:
Use a large enough array to hold the formated public key.

### DIFF
--- a/td-shim/src/secure_boot.rs
+++ b/td-shim/src/secure_boot.rs
@@ -8,9 +8,6 @@
 //! A customized secure boot protocol is designed for td-shim, please refer to `doc/secure_boot.md`
 //! for details.
 
-extern crate alloc;
-
-use alloc::vec::Vec;
 use core::mem::size_of;
 use core::ptr::slice_from_raw_parts;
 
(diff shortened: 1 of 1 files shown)


td-shim/src/secure_boot.rs
```diff
@@ -171,8 +171,9 @@ impl<'a> PayloadVerifier<'a> {
                 signature = &signed_payload[offset..offset + 96];
 
                 // Uncompressed public key
-                formated_public_key.push(0x04);
-                formated_public_key.extend_from_slice(public_key);
+                formated_public_key[0] = 0x04;
+                formated_public_key[1..1 + public_key.len()].copy_from_slice(public_key);
+                public_key_size = 97;
```
jyao1, line 176: why hardcode 97?

td-shim/src/secure_boot.rs
```diff
@@ -45,6 +42,7 @@ pub const CFV_FILE_HEADER_PUBKEY_GUID: Guid = Guid::from_fields(
 
 pub const PUBKEY_FILE_STRUCT_VERSION_V1: u32 = 0x01;
 pub const PUBKEY_HASH_ALGORITHM_SHA384: u64 = 1;
+const ENCODED_PUBKEY_MAX_SIZE: usize = 1024;
```
jyao1, line 45: why hardcode 1024?

td-shim/src/secure_boot.rs
gaojiaqi7: For RSA, the public key input is encoded in DER format.

See https://github.com/briansmith/ring/blob/main/src/rsa.rs#L44-L54

For ECDSA, the public key input is one encoding byte (indicating compression) and two 48 bytes point data:

See https://github.com/briansmith/ring/blob/main/src/ec/suite_b/public_key.rs#L29-L48 and https://github.com/briansmith/ring/blob/main/src/ec/suite_b/ops.rs#L224-L245

jyao1 (approved):

PR #703, opened 2024-05-24
td-shim: use static memory for temp stacks of APs

gaojiaqi7:
Remove the use of heap can improve the functional safety. A piece of large enough static memory is allocated to hold the temparary stacks of APs which can support maximum 15 APs.


td-shim/src/bin/td-shim/td/tdx_mailbox.rs
```diff
@@ -18,6 +18,8 @@ use crate::asm::{ap_relocated_func_addr, ap_relocated_func_size};
 // The maximum size of memory used for AP stacks is 30 KB.
 const MAX_WORKING_AP_COUNT: u32 = 15;
```
jyao1, line 19: I filed https://github.com/confidential-containers/td-shim/issues/705 for clean up.

jyao1 (approved):

PR #704, opened 2024-05-24
td-shim: use fixed-size array for ACPI table installation

gaojiaqi7:
Replace the vector with a fixed-array to record the table offsets of ACPI tables. The maximum number of ACPI tables is 128 according to Linux Kernel implementation.

### DIFF
--- a/td-shim/src/bin/td-shim/acpi.rs
+++ b/td-shim/src/bin/td-shim/acpi.rs
@@ -4,19 +4,20 @@
 // SPDX-License-Identifier: BSD-2-Clause-Patent
 extern crate alloc;
 
-use alloc::vec::Vec;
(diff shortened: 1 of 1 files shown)


td-shim/src/bin/td-shim/acpi.rs
```diff
@@ -4,27 +4,32 @@
 // SPDX-License-Identifier: BSD-2-Clause-Patent
 extern crate alloc;
 
-use alloc::vec::Vec;
 use td_shim_interface::acpi::{calculate_checksum, Rsdp, Xsdt};
 
 use super::*;
 
-#[derive(Default)]
+const ACPI_MAX_TABLES: usize = 128;
```
jyao1, line 11: what if VMM inputs more table number > ACPI_MAX_TABLES?

jyao1 (approved):

Issue #709, opened 2024-06-06
General clean-up proposal to support TD partitioning


vijaydhanraj:
- General clean-ups:
    - Use `addr_of!` macro instead of directly using the `&` operator as we cannot assume the alignment of the reference data (example, tdvmcall_mmio_write/tdvmcall_mmio_write).

PR #711, opened 2024-06-14
Use addr_of! macro instead of &

OuyangHang33:
Fix: #709 

General Clean up.


OuyangHang33:
@vijaydhanraj Could you help to review this PR? Thanks!

jyao1 (approved):

tdx-tdcall/src/tdx.rs
vijaydhanraj, line 1: Thanks Hank! Can addr_of! be used here as well? Otherwise LGTM.
`let buffer: u64 = &digest.data as *const u8 as u64;`

OuyangHang33: Sure, it also need to change. I will update it

vijaydhanraj (approved):
Thanks, LGTM.

Issue #708, opened 2024-06-06
Proposal to change existing TDCALL APIs to support TD Partitioning


vijaydhanraj:
TD partitioning may require few changes to existing TDCALLs. These APIs are listed below,

- Existing API Changes:
    - Add a return type indicating success and failure for tdvmcall_wrmsr, tdvmcall_rdmsr.
    - Extend TdInfo struct to add vcpu_index field.
        - TdInfo stores the result from the tdcall `tdcall_get_td_info`. vcpu_index is one of the values returned in R9 bit0 ~ 31. Save this returned vcpu_index in TdInfo.

PR #712, opened 2024-06-14
Add Result return type for tdvmcall_wrmsr, tdvmcall_rdmsr

OuyangHang33:
Ref: #708 

Replace the` tdvmcall_halt();` with `return Err(ret.into())`

### DIFF
--- a/td-exception/src/interrupt.rs
+++ b/td-exception/src/interrupt.rs
@@ -399,13 +399,15 @@ interrupt_no_error!(virtualization, stack, {
             }
         }
         EXIT_REASON_MSR_READ => {
-            let msr = tdx::tdvmcall_rdmsr(stack.scratch.rcx as u32);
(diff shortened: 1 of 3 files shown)


OuyangHang33:
@vijaydhanraj Could you help to review this pr? thanks!

jyao1 (approved):

td-exception/src/interrupt.rs
vijaydhanraj, line 1: Changes LGTM! Are you planning to create a separate PR to extend `TdInfo struct` with `vcpu_index` field? This is another request as part of https://github.com/confidential-containers/td-shim/issues/708

OuyangHang33: Yes I will submit another PR for this.

vijaydhanraj (approved):

Issue #707, opened 2024-06-05
Add new TDCALL functions to support TD partitioning


vijaydhanraj:
Once https://github.com/confidential-containers/td-shim/issues/706 is done the next step would be to add new TDCALL functions to support TD partitioning. The following new TDCALLs may be required.

- New APIs:
    - Wrapper to `tdcall_accept_page` accept a memory range ( `td_accept_memory` that takes start address and len of address range to be accepted).
    - Add tdcall_vm_read/write to access TD-scope meta field of a TD.
    - Add tdcall_vp_read/write is to access vCPU-scope meta field of a TD.
    - Add tdcall_vp_invept/invvpid to provide SEPT flushing support.
    - Add tdcall_vp_enter support.
    - Add tdcall to support memory attribute write.

jyao1:
@vijaydhanraj, Could you please split the requirement to 3 different categories: 1. Change existing, 2. Clean up, 3. Add new ones?

Usually, we do not put so many request into one issue. The different task be assigned to different owner, with different time.

vijaydhanraj:
Sure, will do accordingly. This issue, I will keep for adding new TDCALLs as the title says and will create two other issues for existing changes and cleanup.

vijaydhanraj:
Created https://github.com/confidential-containers/td-shim/issues/708 for changing existing APIs and created https://github.com/confidential-containers/td-shim/issues/709 for general clean-ups.

PR #714, opened 2024-06-18
Add new TDCALL functions to support TD partitioning

vijaydhanraj:
Fixes: https://github.com/confidential-containers/td-shim/issues/707

Added the following TDCALL support as part of this PR.

- Wrapper to tdcall_accept_page to accept a memory range (for both normal 4K as well as 2M large pages).
- Add tdcall_vm_read/write to access TD-scope meta field of a TD.
- Add tdcall_vp_read/write is to access vCPU-scope meta field of a TD.
- Add tdcall_vp_invept/invvpid to provide SEPT flushing support.
- Add tdcall_vp_enter support.
- Add tdcall to support memory attribute write.


tdx-tdcall/src/asm/tdcall.asm
```diff
@@ -52,22 +53,25 @@ asm_td_call:
         # tdcall
         .byte 0x66,0x0f,0x01,0xcc
 
-        # Exit if tdcall reports failure.
-        test rax, rax
-        jnz td_call_exit
+        # On TDVM exit, RDI loses the RCX value (TdVmcallArgs struct) that was saved earlier.
+        # But r15 register is not used and is cleared to 0 on TDVM exit so use to it get the
+        # TdcallArgs struct stored on the stack.
+        pop r15
```
jyao1, line 59: I am a little confused here. Why not `pop rdi`?

vijaydhanraj: As per the ABI spec, `TDG.VP.ENTER` modifies `rdi` to store the `CS base address`. So preferred to use a register which is not used by the call. 
image

jyao1: That just means RDI is used in TDX module. It does not mean the caller wrapper cannot use it.

See code at line 76, RDI is popped and overridden again.

I agree with you on the old code bug, that RDI is lost and we should not use it directly. But just to pop the saved RCX to RDI is enough. I don't see the need to use R15.

Do you see any failure when you pop RDI?

vijaydhanraj: No, I don't see failure when popping to RDI. I have reverted and use RDI. PTAL.

jyao1 (approved):

PR #724, opened 2024-07-16
tdx-tdcall: fix TdCallError parsing

peterfang:
According to the ABI spec, TdCallError includes:

1) Flags, Class and Name (Bits 63:32)
2) Details L2 (Bits 31:0)

Fix the logic in fn from(val: u64) to correctly parse both parts of the function completion status. Include details L2 as part of the enum.


gaojiaqi7:
Look good from my side

jyao1 (approved):

Issue #720, opened 2024-06-24
td-payload: make private shadow optional


gaojiaqi7:
https://github.com/confidential-containers/td-shim/blob/main/td-payload/src/mm/shared.rs#L15-L30

jyao1:
should be: `init_shared_memory(start: u64, size: usize, shadow_start: Option<u64>)`

PR #730, opened 2024-08-02
td-payload: add new shared memory init function with private shadow

gaojiaqi7:
To keep the original API behavior unchanged. The extended API init the shared memory allocator with a private shadow start address. If the private shadow is not available, the method `copy_to_private_shadow` will return None.

As the `shadow_start` may be lower or higher than start of shared memory, the way of allocating private shadow is changed to use the offset of the allocated shared address to the start of shared allocator.

Closes https://github.com/confidential-containers/td-shim/issues/720


xiaoyuxlu:
LGTM

jyao1 (approved):

Issue #734, opened 2024-08-15
Wrong bitmask on GPAW?


dingelish:
In Flat32ToFlat64.asm I see a weird bitmask `0x2f`. it is `0010 1111` and cannot precisely extract [6:0] bits of gpaw. the correct bitmask should be `0x3f` which is `0011 1111`.

```
    ;
    ; esp [6:0] holds gpaw, if it is at least 52 bits, need to set
    ; LA57 and use 5-level paging
    ;
    mov     ecx, esp
    and     ecx, 0x2f
```

https://github.com/confidential-containers/td-shim/blob/9dea7d2bf0a0bccd621a3a692a2809148f691e96/td-shim/ResetVector/Ia32/Flat32ToFlat64.asm#L25

PR #738, opened 2024-09-11
ResetVector: remove 5-level paging

gaojiaqi7:
5-level paging is not necessary in td-shim, remove the related code and page table entry in reset vector.

As the level 5 entry is removed, update the size of `ResetVector` in image layout.

This PR should close https://github.com/confidential-containers/td-shim/issues/734 as GPAW is no longer needed.

### DIFF
--- a/td-shim/ResetVector/Ia32/Flat32ToFlat64.asm
+++ b/td-shim/ResetVector/Ia32/Flat32ToFlat64.asm
@@ -16,29 +16,11 @@ Transition32FlatTo64Flat:
 
     mov     eax, cr4
     bts     eax, 5                      ; enable PAE
-
-    ;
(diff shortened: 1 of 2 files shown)


devtools/td-layout-config/config_image.json
```diff
@@ -5,6 +5,6 @@
     "TempHeap": "0x020000",
     "Metadata": "0x001000",
     "Payload": "0xC2D000",
-    "Ipl": "0x349000",
-    "ResetVector": "0x008000"
+    "Ipl": "0x34A000",
+    "ResetVector": "0x007000"
```
jyao1, line 9: please dont change size.

jyao1:
Need check CR4.LA57.

jyao1 (approved):

PR #745, opened 2024-10-31
tdx-tdcall: don't use heap allocated buffers for TD reports

Freax13:
There's little reason to use a heap-allocated buffer for requesting TD reports putting it behind a global lock. Furthermore, the TDX module doesn't require the additional data to be inside the TD report buffer. Instead, put the buffer on the stack and directly pass a pointer to the additional data.
Note that with and without this patch, the memory passed to the TDX module must be identity-mapped.

### DIFF
--- a/tdx-tdcall/src/lib.rs
+++ b/tdx-tdcall/src/lib.rs
@@ -21,7 +21,6 @@
 
 #![no_std]
 
-extern crate alloc;
 use core::ffi::c_void;
 
 #[cfg(feature = "use_tdx_emulation")]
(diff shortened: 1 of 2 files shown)


gaojiaqi7:
Thanks @Freax13 . The additional data is required to be 64B-aligned. Maybe we can add it in `TdxReportBuf` or define a new structure with alignment to hold the `additional_data`.

Freax13:
> Thanks @Freax13 . The additional data is required to be 64B-aligned. Maybe we can add it in `TdxReportBuf` or define a new structure with alignment to hold the `additional_data`.

Good catch. Yeah, we can just create a new wrapper struct and copy the data to the stack.

I no longer have the time to work on this PR. Feel free to update or close it.

jyao1 (approved):

Issue #749, opened 2024-12-11
Switch to stable rust toolchain


gaojiaqi7:
There are still some unstable features used by this project, such as `naked_functions`. Dependencies such as `x86_64` rely on nightly toolchain and unstable features.

- [x] Remove the `naked_functions` used in `td-exception` crate
- [x] Clean up the use of `alloc_error_handler` in `td-shim` crate
- [x] Remove dependencies and build tools that rely on nightly toolchain

PR #755, opened 2024-12-18
switch to stable toolchain

gaojiaqi7:
- Remove naked interrupt handler:

An interrupt handler table is used to handle all 256 interrupts. They share common handler that saves the interrupt vector and context. If the interrupt is not an error, a zero is pushed to the stack to align with the error one.

Function `generic_interrupt_handler` is called to find and call the corresponding callback in the `CALLBACK_TABLE`.

For virtualization exception, when the CET shadow stack is enabled, we need to update the LIP value in the shadow stack. As shadow stack saves the latest two return address after interrupt entry and the SSP, the position of saved LIP value is the top address of shadow stack minus 0x18 bytes.

- Disable default-features of `x86_64` crate

Default features of `x86_64` use unstable features of rust that depend on nightly toolchain.

 - Upgrade to latest stable toolchain 1.83.0 and update related scripts and docs.
 - Close compiler warnings.


gaojiaqi7:
Integration test is disabled and tracked by a new issue https://github.com/confidential-containers/td-shim/issues/757

jyao1 (approved):

gaojiaqi7:
Closes https://github.com/confidential-containers/td-shim/issues/749

PR #775, opened 2025-03-13
Bump ring library version to 0.17.14

zlukwins:
Following items included:
* ring library bumping up from 0.17.6 to 0.17.14
* dropping patch for NFC: Address Clippy unused import warning introduced
with https://github.com/confidential-containers/td-shim/pull/760
* dropping SSE and SSE2 checks


zlukwins:
This change is required to fix the ring vulnerability:

image

https://github.com/advisories/GHSA-4p46-pwfr-66x6

library/patches/0001-Support-x86_64-unknown-none-target.patch
```diff
@@ -0,0 +1,119 @@
+From 022337ecdb430a2e536db8dbd3199b9d3e3cf58e Mon Sep 17 00:00:00 2001
```
jyao1, line 1: 1) Can we use original file name - ring.diff?
I dont think we need to change that part.

2) I am not sure about b/src/cpu/intel.rs
Can we just use the feature to disable?
Or if there is issue, I prefer to add another diff.
Please don't mix the function code with build code.

zlukwins: Sure.
Let me address both 1 and 2 ASAP.
Thanks!

zlukwins: Both issues fixed.

jyao1 (approved):

PR #779, opened 2025-03-18
td-exception: release the lock before calling the callback

gaojiaqi7:
When an exception occurs within the interrupt handler, it will re-enter the `generic_interrupt_handler` function. At this point, the `CALLBACK_TABLE` is in a locked state, which can cause the program to fall into a deadlock.

### DIFF
--- a/td-exception/src/interrupt.rs
+++ b/td-exception/src/interrupt.rs
@@ -181,7 +181,11 @@ fn generic_interrupt_handler(stack: &mut InterruptStack) {
         return;
     }
 
-    (CALLBACK_TABLE.lock().table[stack.vector].func)(stack);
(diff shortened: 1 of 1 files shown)


td-exception/src/interrupt.rs
```diff
@@ -181,7 +181,9 @@ fn generic_interrupt_handler(stack: &mut InterruptStack) {
         return;
     }
 
-    (CALLBACK_TABLE.lock().table[stack.vector].func)(stack);
+    // Copy the pointer out, release the lock and then call the function.
```
jyao1, line 184: please add comment:
As such, it can support to handle an exception (#VE) in an interrupt (timer).

jyao1 (approved):

PR #781, opened 2025-03-25
td-shim-tee-info-hash: rework build_mrtd()

mythi:


jyao1:
@mythi , would you mind to provide more background on this patch?

E.g. What problem you want to resolve?
What test you have done to ensure it is still correct?

mythi:
> @mythi , would you mind to provide more background on this patch?
> 
> E.g. What problem you want to resolve?

This is covered in the commit messages.

> What test you have done to ensure it is still correct?

I generated a TD quote and compared its MRTD value with the value computed by the tool and checked that they match.

Xynnn007:
@mythi Any update on this PR? I test the commits but it does not work with the OVMF.fd released from kata deploy v3.23.0 (sha256sum `66923de3388a31fbc1e6f0897d27cee9d8ddcfdb5bf2d26538c6652dd8cbc96d`)

Xynnn007:
Ok I got it worked. My qemu version (ubuntu default) is `8.2.2+ds-0ubuntu1.4+tdx1.1`, and the operation order is firstly ADD all pages and then EXTEND them again due to https://github.com/intel-staging/qemu-tdx/issues/1#issuecomment-2045259345. I did the logic change based on the PR and get the good reference value.

td-shim-tools/src/tee_info_hash.rs
```diff
@@ -320,37 +324,47 @@ impl TdInfoStruct {
...
-                if sec.attributes & TDX_METADATA_ATTRIBUTES_EXTEND_MEM_PAGE_ADD == 0 {
-                    // Use TDCALL [TDH.MEM.PAGE.ADD]
-                    fill_buffer128_with_mem_page_add(
-                        &mut buffer128,
-                        sec.memory_address + iter * PAGE_SIZE,
-                    );
+            let mut section_data = vec![0u8; sec.memory_data_size as usize];
 
-                    sha384hasher.update(buffer128);
+            raw_image_file
```
gaojiaqi7: For certain section types, the `raw data size` might be 0 while the `memory data size` is non-zero. This implies that this memory region should be zero-initialized, and no data needs to be copied from the file.

Attempting to read from the file based on the `memory data size` here is incorrect and can lead to the panic described above.

gaojiaqi7 (changes requested):

td-shim-tools/src/tee_info_hash.rs
mythi: Thanks. Let me handle this differently than just .expect().

mythi: updated

PR #782, opened 2025-04-17
tdx-tdcall: support for new migtd ghci calls

dabron:
Added functions to support new TDVMCALL_MIGTD leaf and subleafs.
Added new tdvmcall which passes RBX (and other registers) to VMM.

### DIFF
--- a/tdx-tdcall/src/asm/mod.rs
+++ b/tdx-tdcall/src/asm/mod.rs
@@ -1,6 +1,7 @@
-// Copyright (c) 2020 Intel Corporation
+// Copyright (c) 2020-2025 Intel Corporation
 //
 // SPDX-License-Identifier: BSD-2-Clause-Patent
+
 use core::arch::global_asm;
 use core::ffi::c_void;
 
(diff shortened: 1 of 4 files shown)


jyao1:
@dabron , would you please add "Signed-off-by: " tag?

tdx-tdcall/src/asm/mod.rs
```diff
@@ -16,8 +17,13 @@ global_asm!(include_str!("tdcall.asm"));
...
 
+#[cfg(all(not(feature = "use_tdx_emulation"), not(feature = "no-tdvmcall")))]
+global_asm!(include_str!("tdvmcall_ex.asm"));
+
 extern "win64" {
     pub(crate) fn asm_td_call(args: *mut c_void) -> u64;
     #[cfg(not(feature = "no-tdvmcall"))]
     pub(crate) fn asm_td_vmcall(args: *mut c_void, do_sti: u64) -> u64;
+    #[cfg(not(feature = "no-tdvmcall"))]
+    pub(crate) fn asm_td_vmcall_ex(args: *mut c_void) -> u64;
```
jyao1, line 28: I think we should keep `do_sti` parameter, from `asm_td_vmcall`

tdx-tdcall/src/tdx.rs
```diff
@@ -513,6 +513,145 @@ pub fn tdvmcall_service(
...
+    }
+
+    Ok(())
+}
+
+#[cfg(not(feature = "no-tdvmcall"))]
+pub fn tdvmcall_migtd_reportstatus(
+    mig_request_id: u64,
+    pre_migration_status: u8,
+) -> Result<(), TdVmcallError> {
```
jyao1, line 556: We also need 
```
    data_buffer: &mut [u8],
    interrupt: u8,
```

jyao1 (approved):

PR #788, opened 2025-05-20
td-shim-tools: add igvm support

dabron:
Add an optional command-line parameter "--image-format" with possible values of "tdvf" and "igvm". The default is "tdvf". If "igvm" is specified, then the output image will be generated in the IGVM format. See https://crates.io/crates/igvm_defs.


bhargavshah1988:
@dabron Shall we consider upating the MigTd Readme about this new build option ?

PR #789, opened 2025-06-02
allow customizable fw location

dabron:
Add an optional command-line parameter: "--memory-size" to specify the GPA where FW will be loaded beneath. The default value is 0x100000000 (4GB). This allows td-shim FW to be loaded to a specified address rather than always beneath 4GB. The reset vector page will always be loaded at 0xfffff000 since that is architectural.


td-shim/src/bin/td-shim/memory.rs
```diff
@@ -52,7 +54,7 @@ impl<'a> Memory<'a> {
             .iter()
             .map(|entry| {
                 let entry_top = entry.physical_start + entry.resource_length;
-                if entry.resource_type == RESOURCE_SYSTEM_MEMORY && entry_top < MEMORY_4G {
+                if entry.resource_type == RESOURCE_SYSTEM_MEMORY && entry_top < MEMORY_SIZE {
```
jyao1, line 57: why change this?
I think that firmware size should be irrelevant to MEMORY_4G.

td-shim/src/bin/td-shim/memory.rs
```diff
@@ -190,8 +192,8 @@ impl<'a> Memory<'a> {
             // For example, QEMU may reserve 4 pages at 0xfeffc000 for an EPT
             // identity map and a TSS in order to use vm86 mode to emulate
             // 16-bit code directly.
-            if new.physical_start >= MEMORY_4G - RESERVED_MEMORY_SPACE_SIZE
-                && new.physical_start + new.resource_length < MEMORY_4G
+            if new.physical_start >= MEMORY_SIZE - RESERVED_MEMORY_SPACE_SIZE
+                && new.physical_start + new.resource_length < MEMORY_SIZE
```
jyao1, line 196: why change this?
I think that firmware size should be irrelevant to MEMORY_4G.

td-shim/src/bin/td-shim/memory.rs
```diff
@@ -207,14 +209,15 @@ impl<'a> Memory<'a> {
...
             {
-                if entry_top > MEMORY_4G {
+                if entry_top > MEMORY_SIZE {
                     if new.resource_type == RESOURCE_SYSTEM_MEMORY
                         || new.resource_type == RESOURCE_MEMORY_UNACCEPTED
                     {
-                        new.physical_start = MEMORY_4G;
-                        new.resource_length = entry_top - MEMORY_4G;
+                        new.physical_start = MEMORY_SIZE;
+                        new.resource_length = entry_top - MEMORY_SIZE;
```
jyao1, line 220: why change this?
I think that firmware size should be irrelevant to MEMORY_4G.

jyao1:
Maybe I am not clear, if we move the fw location, how do we handle the memory space between [fw.base + fw.size, 4G] ? Usable? Reserved?

td-shim/src/bin/td-shim/memory.rs
dabron: Removed unnecessary change.

td-shim/src/bin/td-shim/memory.rs
dabron: Removed unnecessary change.

dabron:
> Please also share how you do the test to ensure the code is still correct, if people move the FW to any arbitrary location.
> 
> Or if there is any restriction on the FW location?
> 
> For example: What if TD-SHIM is put to [248MB, 256MB], while the whole memory is 2G? Is that a valid configuration? If it is valid, then is the memory between [256MB, 2G] usable for this TD? Or is the TD can only use the memory between [0, 248MB] ?
> 
> And what if TD-SHIM is put to [56MB, 64MB], or even [8MB, 16MB], [1MB, 9MB] ?
> 
> Have you validated such configuration?

The TD will only not use any memory above the TD-SHIM range. So when using IGVM, MEMORY_SIZE needs to include system memory size + payload size (e.g. MigTD requires 32MB + 16MB = 48MB).

jyao1:
A) 8G: [0, 3G] [4G, 9G] - 1) reported in TD_HOB, similar to TDVF, 2) it is recorded in TD-metadata table, used by MigTD/TPA.
B) 8G: [0, 3G-FW_size] [4G, 9G] - 1) reported in TD_HOB, similar to TDVF, 2) it is recorded in TD-metadata table, used by MigTD/TPA.
NOTE: This patch only handles 2). The 1) is not supported, and new configuration will be ignore.
Question: which one is in metadata table? VMM need to provide A) and put FW into right place: FW_BASE = 3G - FW_SIZE.

FW location: [3G-FW_Size, 3G] - fixed at build time. (no dynamic at this time, although it could be preferred).
Question: how VMM knows the FW location? IGVM structure includes this information.
When input the size, do you prefer 3G-FW_Size or 3G? Currently, it is 3G-fw_Size. Maybe we should call "fw_Base" ?

What if :8G: [0, 2G] [4G, 10G]? What will happen? - cause TD fail at reset vector.
What if :8G: [0, 3.5G] [4G, 10G]? What will happen? - [3G, 3.5G] will NOT be accepted or used. TD-SHIM will report in memory map.

How does TD-shim report [3G-FW_Size, 3G] in memory map? Maybe it is usable today. Then it is an issue. It should be reported as reserved_memory.

Finally, we need a new document to tell how to make IGVM work.

jyao1:
> So if the container size is more then 4 GB then, [fw.base + fw.size, 4G] is a usable memory. MigTd can use it. If the container size is less then 4 GB then, there will be a gap in address space layout. So its not usable memory.
> 
> @jyao1 Does this answer your question ?

Yes, you answered my question.

However, I am not sure why we design in different way.
Why cannot we use same rule to make [fw.base + fw.size, 4G] to be usable or unusable in both cases?

dabron:
> > So if the container size is more then 4 GB then, [fw.base + fw.size, 4G] is a usable memory. MigTd can use it. If the container size is less then 4 GB then, there will be a gap in address space layout. So its not usable memory.
> > @jyao1 Does this answer your question ?
> 
> Yes, you answered my question.
> 
> However, I am not sure why we design in different way. Why cannot we use same rule to make [fw.base + fw.size, 4G] to be usable or unusable in both cases?

I don't think that there is a different design. If container size is less than 4G, then [fw.base + fw.size, container size] would be usable and [container size, 4G] would not.

jyao1:
Question: Is there any unit test to cover all those different combinations?

jyao1:
> > > So if the container size is more then 4 GB then, [fw.base + fw.size, 4G] is a usable memory. MigTd can use it. If the container size is less then 4 GB then, there will be a gap in address space layout. So its not usable memory.
> > > @jyao1 Does this answer your question ?
> > 
> > 
> > Yes, you answered my question.
> > However, I am not sure why we design in different way. Why cannot we use same rule to make [fw.base + fw.size, 4G] to be usable or unusable in both cases?
> 
> I don't think that there is a different design. If container size is less than 4G, then [fw.base + fw.size, container size] would be usable and [container size, 4G] would not.

@bhargavshah1988 , please confirm your understanding.

bhargavshah1988:
> > > > So if the container size is more then 4 GB then, [fw.base + fw.size, 4G] is a usable memory. MigTd can use it. If the container size is less then 4 GB then, there will be a gap in address space layout. So its not usable memory.
> > > > @jyao1 Does this answer your question ?
> > > 
> > > 
> > > Yes, you answered my question.
> > > However, I am not sure why we design in different way. Why cannot we use same rule to make [fw.base + fw.size, 4G] to be usable or unusable in both cases?
> > 
> > 
> > I don't think that there is a different design. If container size is less than 4G, then [fw.base + fw.size, container size] would be usable and [container size, 4G] would not.
> 
> @bhargavshah1988 , please confirm your understanding.

Yes, this is correct understandings.

dabron:
> Question: Is there any unit test to cover all those different combinations?
> 
> Also the documentation?

Added documentation on how to build IGVM image, also build tests for the new command line arguments.

dabron:
> 1. Where can I find the document to describe the use case? E.g. It only support metadata based use case. It does not support TD_HOB as input.
> 2. Where can I find the document to describe the memory layout? E.g.
> 
> * fw.top = (fw.base + fw.size) is expected to set to LowTop (LowTop means Top memory below 4GiB)
> * [0, fw.base] is always usable memory.
> * [fw.base + fw.size, LowTop] is usable memory, if it is present.
> * [4G, HighTop] is always usable memory. (HighTop means Top memory above 4GiB)

Added igvm_image.md

dabron:
> BTW: I just think one thing: Why we need end user input the fw.top (== fw.base + fw.size), instead of setting fw.top == LowTop automatically? What is the use case to set fw.top < LowTop and leave a small window there?

We need the fw_top value when calling td-layout-config -t image. Do we know the low_top value at this point?

jyao1:
>> BTW: I just think one thing: Why we need end user input the fw.top (== fw.base + fw.size), instead of setting fw.top == LowTop automatically? What is the use case to set fw.top < LowTop and leave a small window there?

> We need the fw_top value when calling td-layout-config -t image. Do we know the low_top value at this point?

According to https://github.com/confidential-containers/td-shim/pull/789#issuecomment-3034024260, this patch only supports **TD-metadata** use case, and it does not support **TD-hob** use case.

It means that, given the **TD-metadata** config, you already knows the low_top. I don't see the need to let use input something else.
What if the TD-metadata indicates to support 1G, while user made mistake to input 2G?

The ground rule for tool is: if it can be calculated by itself, then don't ask for user input.

PR #810, opened 2025-11-13
td-shim-tools: unmeasure CFV for igvm

haitaohuang:
CFV memory should not be in MRTD otherwise we have circular dependency between policy (which is in CFV) and MRTD measurement. CFV memories are measured in RTMRs.


haitaohuang:
discussed with Jiewen, looks like there are other cases we want to measure. Will create an issue

PR #845, opened 2026-05-05
td-layout-config: add TempMemBase to relocate TempMem entries to guest RAM

MichalTarnacki:
QEMU's tdx_accept_ram_range() only searches E820 RAM entries when accepting TDX memory. TempMem metadata sections with addresses in the firmware pflash range (0xFF000000+) cause 'Failed to accept memory for TDVF section' errors on upstream QEMU without patches.

Add optional "TempMemBase" field to image_layout.json. When set, td-layout-config relocates the _BASE addresses of Mailbox, TempStack, and TempHeap entries from the firmware range into the specified guest RAM address, making them acceptable by unpatched QEMU.

Changes:
- layout.rs: add base_override field to LayoutEntry, add relocate_entries() method
- image.rs: parse TempMemBase from config, call relocate_entries()
- image.jinja: emit base_override when present instead of memory_offset + region.start

The feature is opt-in and backward compatible: without TempMemBase in the JSON config, behavior is unchanged.

fixes #834

### DIFF
--- a/devtools/td-layout-config/config_image.json
+++ b/devtools/td-layout-config/config_image.json
@@ -6,5 +6,6 @@
     "Metadata": "0x001000",
     "Payload": "0xC2D000",
     "Ipl": "0x349000",
-    "ResetVector": "0x008000"
+    "ResetVector": "0x008000",
+    "TempMemBase": "0x1FBF000"
 }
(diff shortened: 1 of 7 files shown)


jyao1:
please fix CI failure

PR #846, opened 2026-05-11
td-layout: place TempMem sections in DRAM for QEMU 9+ TDX compatibility

hyjiang:
## Problem

QEMU 9.x and later changed how TDX private memory is initialized. The
function `tdx_init_ram_entries()` now only populates TDX private memory
from E820_RAM entries.  The three TempMem TDVF sections (Mailbox,
TempStack, TempHeap) were located in the firmware ROM region
(0xFF040000–0xFF081000), which is not covered by E820_RAM.  As a
result, QEMU rejects those sections with:

```
Failed to accept memory for TDVF section 3
```

## Solution

Add an optional `TempMemBase` field to `config_image.json`.  When
present, the layout generator places Mailbox, TempStack, and TempHeap
at a DRAM address instead of deriving them from the ROM firmware offset.
The generated `build_time.rs` constants then point into DRAM (e.g.
0x600000), which is within the Bootloader E820_RAM region and therefore
accepted by QEMU's TDX initialisation.

Because the field is optional, omitting it preserves the previous
(ROM-based) behaviour, keeping backward compatibility with older QEMU
versions that do not require E820_RAM backing.

## Changes

* `devtools/td-layout-config/config_image.json` – add `TempMemBase`
* `devtools/td-layout-config/src/image.rs` – parse `TempMemBase`
* `devtools/td-layout-config/src/layout.rs` – add `dram_base` field
  and `set_tempmem_dram_base()` helper
* `devtools/td-layout-config/src/template/image.jinja` – emit DRAM
  address when `dram_base` is set
* `td-layout/src/build_time.rs` – regenerated

## Testing

Built with `cargo image --example-payload` and verified with
`td-shim-checker` that TempMem sections are at 0x600000/0x601000/0x621000.
Booted in QEMU 10.2.1 with TDX hardware without
the section-accept error.

### DIFF
--- a/devtools/td-layout-config/config_image.json
+++ b/devtools/td-layout-config/config_image.json
@@ -1,5 +1,6 @@
 {
     "Config": "0x040000",
+    "TempMemBase": "0x600000",
     "Mailbox": "0x001000",
     "TempStack": "0x020000",
     "TempHeap": "0x020000",
--- a/devtools/td-layout-config/src/image.rs
+++ b/devtools/td-layout-config/src/image.rs
@@ -30,6 +30,14 @@ struct ImageConfig {
     reset_vector: String,
     #[serde(rename = "ImageSize")]
     image_size: Option<String>,
+    /// DRAM base address for TempMem sections (Mailbox, TempStack, TempHeap).
+    ///
+    /// When set, these sections are placed at DRAM addresses (starting from this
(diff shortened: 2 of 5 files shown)


MichalTarnacki:
This pr addresses the same issue as #845 but with a bit different approach. Maybe this PR shows a better solution

PR #852, opened 2026-05-28
docs: add README files for all crates without a README

BFuhry:
At the moment, many Rust crates do not have a corresponding README making the repository hard to understand.
I used AI to create initial README files, which helped me a lot in understanding the structure.
After some adjustments to the created files, I now contribute the READMEs to the repo to help others.
Obviously, feel free to adjust the content and in particular you might consider to drop pieces that you consider too verbose.


jyao1:
@BFuhry , thanks for the (AI) work.

Have you verified by yourself if the content AI generated is correct?

If there is any code / command generated, have you run it by yourself to ensure they are correct and other people can just them?

BFuhry:
As I only wanted to understand the repo on a high level, I did not run any piece of it.
So the short answer is: no, I did not verify and code / command generated.

For me, the main value of this PR is the overview text for each crate. Without such a text, a user of the repo is totally lost. Still, I cannot guarantee the correctness as I did not write the code and did not look into it in detail. However, I assume it should be rather quick for any maintainer of this project to go through the text and fix incorrect details.

Regarding code / commands in the README:
I would consider this really valuable for customers, but I agree that these have to be thoroughly validated (and kept up to date).
As such, I'm totally fine to remove all code / commands from this PR, as I will not be able to do the validation myself and I understand that you might consider this too much work.
This is what I meant by "drop pieces that you consider too verbose" in the PR message.

jyao1:
> Regarding code / commands in the README:
I would consider this really valuable for customers, but I agree that these have to be thoroughly validated (and kept up to date).
As such, I'm totally fine to remove all code / commands from this PR, as I will not be able to do the validation myself and I understand that you might consider this too much work.
This is what I meant by "drop pieces that you consider too verbose" in the PR message.

I am not worried about "too verbose". What I am worried about is "correctness".

To me, if there is no way to know if the message is correct, I would rather NOT add it.

Please only commit the message you have verified.

If you do not want to verify and just take whatever AI told you, then I prefer to drop it.
Other people can ask AI as well ...

BFuhry:
@sgrams

PR #856, opened 2026-07-28
td-payload: gate apic import behind no-interrupt feature

sgrams:
Gate the apic import with feature flag to match its usage site, fixing an unused import warning when the no-interrupt feature is enabled.


sgrams:
deny fails due to vulnerable spin package, see https://github.com/confidential-containers/td-shim/pull/857

jyao1 (approved):

